PR #4, opened 2020-05-18
Add verification and testing targets for CI

ixdy:
This PR updates `make test` to use gotestsum (for better test reporting), including automatically generating junit when run in the Prow environment.

It also adds several verification checks, wrapped by the `make verify` rule:
- verify-boilerplate, imported from repo-infra (without using the bazel target)
- verify-lint, which uses `golangci-lint` (with possibly too many linters enabled; further tuning may be necessary)
- verify-modules, which ensures that the go modules files are up-to-date.

After this is in, we should be able to create one or more Prow jobs that run `make build`, `make test`, and `make verify`.

/assign @alvaroaleman


hack/verify/verify_modules.sh
```diff
@@ -0,0 +1,39 @@
...
+    fi
+}
+
+trap cleanup EXIT
+
+cp -a "${REPO_ROOT}" "${TMP_REPO}"
+cd "${TMP_REPO}/boskos"
+make update-modules >/dev/null
+
+if !(git diff --quiet HEAD -- go.sum go.mod hack/tools/go.mod hack/tools/go.sum); then
```
alvaroaleman, line 36: I think the leading `!` is wrong?

Makefile
```diff
@@ -12,25 +12,70 @@
...
 # TODO(ixdy): containerize
+.PHONY: build
 build:
 	go build ./...
 
-test:
-	go test ./...
+.PHONY: test
+test: $(GOTESTSUM)
+	$(GOTESTSUM) $${ARTIFACTS:+--junitfile="${ARTIFACTS}/junit.xml"} ./...
```
alvaroaleman, line 34: TILd

hack/verify/verify_modules.sh
ixdy: the parens were weird (I basically copied this from a Makefile) so I've now removed them, but otherwise this should do the right thing - `git` exits 0 if no diff, nonzero if diff, so we want to invert that.

Makefile
ixdy: this is maybe too clever, but it works for now...

alvaroaleman:
/lgtm
We should activate that the github review can be used to approve

PR #5, opened 2020-05-19
verify_boilerplate: don't require python3

ixdy:
It seems like this script isn't using any python3-specific features, and the default golang image doesn't have python3 (which makes running this in CI more annoying).


ixdy:
whoops, this doesn't actually work...

PR #6, opened 2020-05-19
Fix verify_boilerplate to work in python2

ixdy:
take 2 (still testing locally)


ixdy:
This seems to work now. `io.open()` and `open()` are apparently the same thing in python3. In python2 it seems like you can use `io.open()` to get similar functionality, hence this should work.

PR #8, opened 2020-05-21
Build images without bazel

ixdy:
I mostly tried to keep the same functionality from the Bazel builds, but I made a few changes:
- I renamed the `cli` image to `boskosctl`, since its referred to as boskosctl in all of the docs, and so seems more logical, but I can change that back. (@stevekuznetsov any thoughts?) I couldn't tell if anyone was using the existing `gcr.io/k8s-prow/boskos/cli` image.
- I switched the aws-janitor images to distroless (from gcloud-in-go), since as far as I can tell, they don't need to shell out or call gcloud.
- I've updated to explicitly require go1.14.3 (test-infra was most recently on go1.13.x). I don't think there's any harm in upgrading?
- I've created a cloudbuild configuration for building images, since I think that's the recommended approach from the k8s.io infra working group. It should be easy to connect this to a prow job on postsubmit. (I've tested this locally using `cloud-build-local`.)

I'll remove Bazel in a follow-up PR. This one was already getting a bit large.

/assign @alvaroaleman


alvaroaleman:
>I renamed the cli image to boskosctl, since its referred to as boskosctl in all of the docs, and so seems more logical, but I can change that back. (@stevekuznetsov any thoughts?) I couldn't tell if anyone was using the existing gcr.io/k8s-prow/boskos/cli image.

We mirror it but do not seem to actually be using it so thats fine. Not sure who else might be using it.
/lgtm

PR #9, opened 2020-05-21
Switch to gcr.io/k8s-staging-boskos

ixdy:
Rather than using `gcr.io/k8s-prow`, we should probably use our own staging project.

I've created https://github.com/kubernetes/k8s.io/pull/890 to initiate this process. We shouldn't merge this PR until it's in, probably.

/hold

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 WHAT ?= ./...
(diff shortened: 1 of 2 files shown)


ixdy:
/hold cancel

Seems like we're unblocked on creating staging projects.

PR #24, opened 2020-06-10
Stamp binaries with component name and version and update logrus dependency

### LINKED ISSUE #22: Logger is not configured correctly
Looking at logs from a recent boskos build (from this repo) reveals that some things are not configured appropriately:

boskos:
```
{"component":"unset","file":"/go/pkg/mod/github.com/sirupsen/logrus@v1.4.2/entry.go:314","func":"github.com/sirupsen/logrus.(*Entry).Logf","handler":"handleUpdate","level":"info","msg":"From 10.44.2.234:42282","time":"2020-06-05T23:01:09Z"}
```

cleaner:
```
{"component":"unset","file":"/go/pkg/mod/github.com/sirupsen/logrus@v1.4.2/logger.go:192","func":"github.com/sirupsen/logrus.(*Logger).Log","level":"info","msg":"Cleaner started","time":"2020-06-05T22:47:27Z"}
```

etc.

The component is unset because we aren't setting the variables in k8s.io/test-infra/prow/version.
We can fairly easily add linker flags for this, though it'll require a bit of work in the `Makefile` (or maybe we'll just want to write a wrapper script).

I'm not sure why we're getting useless file and function annotations, though.

/kind bug

ixdy:
Additionally, this has the side effect of building an image for `aws-resources-list`.

Fixes #22.

/assign @alvaroaleman

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,8 @@ OUTPUT_BIN_DIR := $(OUTPUT_DIR)/$(BIN_DIR)
 GOTESTSUM := $(TOOLS_BIN_DIR)/gotestsum
 GOLANGCI_LINT := $(TOOLS_BIN_DIR)/golangci-lint
 
+CMDS = $(notdir $(shell find ./cmd/ -maxdepth 1 -type d | sort))
+
 export GO_VERSION=1.14.3
 export GO111MODULE=on
 export DOCKER_REPO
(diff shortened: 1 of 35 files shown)


ixdy:
I pushed one more commit standardizing on logrus everywhere (including aws-janitor bits).

@alvaroaleman can you take a look when you get a chance? thanks!

PR #26, opened 2020-06-17
Update k8s.io/test-infra dependency and refactor some flags handling

ixdy:
Mostly this is for https://github.com/kubernetes/test-infra/pull/17976.

### DIFF
--- a/cmd/boskos/boskos.go
+++ b/cmd/boskos/boskos.go
@@ -29,7 +29,9 @@ import (
 	"github.com/spf13/viper"
 	corev1 "k8s.io/api/core/v1"
 
+	"k8s.io/test-infra/pkg/flagutil"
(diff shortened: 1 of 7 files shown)


ixdy:
/hold cancel

I hadn't quite updated test-infra to latest, somehow.

ixdy:
OK, looks good now. Of course that wasn't as easy as I'd intended:

- https://github.com/kubernetes/test-infra/pull/17772 changed the `ServePProf` and `ExposeMetrics` functions to require ports
- I opted to use the `InstrumentationOptions` added in that change as well
- To reduce code slightly, I modified the `KubeClientOptions` to implement the test-infra `OptionGroup` interface
- I added error-checking for the `Validate` call which was missing.

PR #28, opened 2020-07-10
Create an opinionated Boskos deployment using kustomize

ixdy:
I know we haven't really discussed exactly how we want deployments to work long-term, and I'm not sure how this integrates with bazel-based deployment systems (such as kubernetes/test-infra), but I think that using kustomize generally makes things pretty clear to understand. I guess you can always run `kustomize` and save the output into static YAML files.

I took the existing manifests from kubernetes/test-infra, plus a few others, and massaged them a bit to simplify them.
I haven't yet added the AWS janitor into the mix, though that should be a fairly straightforward followup.
I also didn't enable v2 of the cleaner, since that isn't widely deployed, and it'll probably be helpful to see the diff, anyway.

I've tested this on a kind cluster with `kustomize build deployments/overlays/example  | kubectl apply -f-`, so that's pretty neat. (The kustomize library linked into `kubectl` seems to barf on the inline patches for some reason.)

/assign @alvaroaleman 
cc @chemikadze

### DIFF
--- a/deployments/base/cleaner/deployment.yaml
+++ b/deployments/base/cleaner/deployment.yaml
@@ -1,0 +1,44 @@
+# Copyright 2020 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: boskos-cleaner
+  labels:
+    app: boskos-cleaner
+  namespace: boskos
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: boskos-cleaner
+  template:
+    metadata:
+      labels:
+        app: boskos-cleaner
+    spec:
+      serviceAccountName: boskos-cleaner
+      terminationGracePeriodSeconds: 300
+      containers:
+        - name: boskos-cleaner
+          image: gcr.io/k8s-staging-boskos/cleaner:v20200617-f289ba6
+          args:
+            - --boskos-url=http://boskos
+            - --namespace=$(NAMESPACE)
+          env:
+            - name: NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
--- a/deployments/base/cleaner/deployment.yaml
+++ b/deployments/base/cleaner/deployment.yaml
@@ -37,6 +37,7 @@ spec:
           args:
             - --boskos-url=http://boskos
             - --namespace=$(NAMESPACE)
+            - --use-v2-implementation
           env:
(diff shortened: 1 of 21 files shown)


deployments/base/cleaner/deployment.yaml
```diff
@@ -0,0 +1,39 @@
...
+        app: boskos-cleaner
+    spec:
+      serviceAccountName: boskos-cleaner
+      terminationGracePeriodSeconds: 300
+      containers:
+        - name: boskos-cleaner
+          image: gcr.io/k8s-staging-boskos/cleaner:v20200617-f289ba6
+          args:
+            - --boskos-url=http://$(BOSKOS_SERVICE).$(BOSKOS_NAMESPACE).svc.cluster.local.
+            - --namespace=$(BOSKOS_NAMESPACE)
```
alvaroaleman, line 39: FWIW, we use the controller-based implementation since February 27th and haven't noticed any issue so far: https://github.com/openshift/release/pull/7347

deployments/base/kustomization.yaml
```diff
@@ -0,0 +1,42 @@
...
+commonLabels:
+  app: boskos
+
+resources:
+  - crd.yaml
+  - deployment.yaml
+  - rbac.yaml
+  - service.yaml
+
+vars:
```
alvaroaleman, line 27: Why do we define this here? To not repeat ourselves?

deployments/base/cleaner/deployment.yaml
ixdy: Thanks, I've added an additional commit so the changes needed are obvious.

deployments/base/kustomization.yaml
ixdy: I initially did this so you could override the namespace and have the command-line arguments updated appropriately, but looking at the openshift examples, using downward API is even easier, so I've switched to that.

ixdy:
Thanks for the review! I've also updated the README.md in the deployments/overlays/example directory showing the commands to use this.

deployments/overlays/example/README.md
```diff
@@ -0,0 +1,26 @@
...
+- `manual-token`
+- `automatic-token` (automatically scaled using Dynamic Resource Lifecycles, though no Mason implementation is used here)
+
+Beyond the core Boskos server, the following components are installed:
+- The `cleaner`, to clean up `automatic-token`s
+- The `reaper`, to clean up orphaned leases
+- Two deployments of the `janitor`, one to clean up `numeric` projects, the other to clean up `phonetic` projects. Note that we increased the number of replicas for the latter.
+  - In this toy example we overrode the real janitor binary to always pass, since these are not real projects and we do not have a real service account. A real implementation would need to ensure the janitor service account has the necessary credentials, for example by using [Workload Identity on GKE](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity).
+
+To try out this example, you will need to [install Kustomize](https://kubernetes-sigs.github.io/kustomize/installation/). To play with it locally, you first create a [kind cluster](https://kind.sigs.k8s.io/).
```
alvaroaleman, line 15: Shouldn't it also work via `kubectl apply -k`?

ixdy: kubectl is using [a rather old version of kustomize](https://github.com/kubernetes/kubernetes/issues/82905) which, among other things, doesn't support inline patches in `kustomization.yaml`.

alvaroaleman:
/lgtm
Thanks!

ixdy:
added a note explaining that. I think you might not need to lgtm again, though.

PR #31, opened 2020-07-10
gcp-janitor: update list of GCP zones

ixdy:
This was last updated a few months ago, and there are more zones now.

I think this is probably safer than relying on `gcloud` to do the right thing itself, given outages we've had in the past. Long-term we still plan to rewrite this janitor in Go.

### DIFF
--- a/cmd/janitor/gcp_janitor.py
+++ b/cmd/janitor/gcp_janitor.py
@@ -72,6 +72,82 @@ DEMOLISH_ORDER = [
     Resource('', 'logging', 'sinks', None, None, None, False, False),
 ]
 
+# gcloud compute zones list --format="value(name)" | sort | awk '{print "    \x27"$1"\x27," }'
+ZONES = [
+    'asia-east1-a',
+    'asia-east1-b',
+    'asia-east1-c',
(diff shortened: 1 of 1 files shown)


cmd/janitor/gcp_janitor.py
```diff
@@ -72,6 +72,82 @@
     Resource('', 'logging', 'sinks', None, None, None, False, False),
 ]
 
+# gcloud compute zones list --format="value(name)" | sort | awk '{print "    \x27"$1"\x27," }'
+ZONES = [
```
chemikadze, line 76: By the way, what is the reason for this to be defined liked that, and not be resolved in runtime via API call?

chemikadze: Ah, just saw that comment about outages, disregard that.

ixdy: We've had issues in the past where the list of all zones computed by gcloud itself (likely using this same API call, though I'm not sure) doesn't match the list of zones known to things like the compute endpoint, which causes it to fail.

ixdy:
definitely need to fix that flake...
/retest

PR #38, opened 2020-08-19
GCP janitor: don't attempt to delete undeletable logging sinks

### LINKED ISSUE #37: GCP janitor failing when trying to clean up logging sinks
At some point, it seems like GCP added two new Cloud Logging logs router sinks to projects:
- `_Default`
- `_Required`

These [cannot be deleted](https://cloud.google.com/logging/docs/export/configure_export_v2), and this recently started causing cleanup of projects to fail, with error messages like the following:
```
ERROR: (gcloud.logging.sinks.delete) PERMISSION_DENIED: Sink _Default cannot be deleted. Consider disabling instead
ERROR: (gcloud.logging.sinks.delete) PERMISSION_DENIED: Sink _Required cannot be deleted
Error try to delete resources sinks: CalledProcessError()
Error try to delete resources sinks: CalledProcessError()
```

ixdy:
This change adds an additional attribute to Resources in the GCP janitor, allowing specifying a list of names that should never be deleted.

This functionality is used to ensure that the `_Default` and `_Required` logging sinks are never deleted.

(More hacks, my favorite!)

Fixes #37

### DIFF
--- a/cmd/janitor/gcp_janitor.py
+++ b/cmd/janitor/gcp_janitor.py
@@ -26,58 +26,59 @@ import sys
 import threading
 
 # A resource that need to be cleared.
+# Any names in preserved_names will never be deleted.
 Resource = collections.namedtuple(
-    'Resource', 'api_version group name subgroup condition managed tolerate bulk_delete')
+    'Resource', 'api_version group name subgroup condition managed tolerate bulk_delete preserved_names')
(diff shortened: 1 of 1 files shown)


ixdy:
I ran this script locally in `--dryrun` mode and it appeared to do the right thing.

PR #43, opened 2020-10-22
support boskos resource of type image-builder-aws-account

kkeshavamurthy:
Changes related to utilizing boskos for image-builder CI, 
https://github.com/kubernetes/test-infra/pull/19645

Please let me know if I have missed anything. 
CC: @detiber

### DIFF
--- a/cmd/aws-janitor-boskos/main.go
+++ b/cmd/aws-janitor-boskos/main.go
@@ -35,6 +35,7 @@ import (
 
 var (
 	boskosURL          = flag.String("boskos-url", "http://boskos", "Boskos URL")
+	rTypes             common.CommaSeparatedStrings
 	username           = flag.String("username", "", "Username used to access the Boskos server")
 	passwordFile       = flag.String("password-file", "", "The path to password file used to access the Boskos server")
 	region             = flag.String("region", "", "The region to clean (otherwise defaults to all regions)")
@@ -47,6 +48,10 @@ var (
 const (
 	sleepTime = time.Minute
 )
+
+func init() {
+	flag.Var(&rTypes, "resource-type", "comma-separated list of resources need to be cleaned up")
+}
 
 func main() {
 	logrusutil.ComponentInit()
@@ -64,6 +69,11 @@ func main() {
 		sweepSleepDuration = d
 	}
 
+	if len(rTypes) == 0 {
+		logrus.Info("--resource-type is empty! Setting it to default: aws-account")
+		rTypes = []string{"aws-account"}
+	}
+
 	boskos, err := client.NewClient("AWSJanitor", *boskosURL, *username, *passwordFile)
 	if err != nil {
 		logrus.WithError(err).Fatal("unable to create a Boskos client")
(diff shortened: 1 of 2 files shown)


common/aws/common.go
```diff
@@ -25,22 +25,38 @@ import (
...
-
 	// UserDataAccessIDKey is the key in UserData containing the AWS Access key ID
 	UserDataAccessIDKey = "access-key-id"
 
 	// UserDataSecretAccessKey is the key in UserData containing the AWS Secret access key
 	UserDataSecretAccessKey = "secret-access-key"
 )
 
+func GetSupportedAWSBoskosResourceTypes() []string {
+	// Supported Resource Types for Boskos AWS accounts
```
ixdy, line 36: rather than hard-coding this list in here, maybe it'd be better to pass a comma-delimited list of resources types as a flag? that way we wouldn't need any additional changes if another aws-account resource type was added.

common/aws/common.go
```diff
@@ -25,22 +25,38 @@ import (
...
+	return []string{
+		"aws-account",
+		"image-builder-aws-account",
+	}
+}
+
+func checkResourceType(resource string, supportedResourceTypes []string) bool {
+	for _, supportedResource := range supportedResourceTypes {
```
ixdy, line 44: maybe we could require the resource type to end with `aws-account` as an invariant instead?

ixdy (changes requested):
/ok-to-test

common/aws/common.go
detiber: ooh, I like that idea

cmd/aws-janitor-boskos/main.go
```diff
@@ -64,6 +70,10 @@ func main() {
 		sweepSleepDuration = d
 	}
 
+	if len(rTypes) == 0 {
+		logrus.Fatal("--resource-type must not be empty!")
```
detiber, line 74: If you add a default for `aws-account` it will avoid needing to coordinate a kubernetes/test-infra PR to prevent from breaking the existing configuration for the aws janitor when this merges.

kkeshavamurthy: Agree. Thank you.

cmd/aws-janitor-boskos/main.go
```diff
@@ -35,6 +36,7 @@ import (
 
 var (
 	boskosURL          = flag.String("boskos-url", "http://boskos", "Boskos URL")
+	rTypes             common.CommaSeparatedStrings
```
ixdy, line 39: maybe set the default value here instead?

kkeshavamurthy: When I set the default value here, I hit an error due to [this] check(https://github.com/kubernetes-sigs/boskos/blob/ae3497d24cd75fef284c9d3c766d63a6b2416a58/common/common.go#L211)

```
./_output/bin/aws-janitor-boskos -resource-type=aws-account,abc-aws-account
invalid value "aws-account,abc-aws-account" for flag -resource-type: resTypes flag already set
```

cmd/aws-janitor-boskos/main.go
```diff
@@ -74,22 +85,25 @@ func main() {
 }
 
 func run(boskos *client.Client) error {
+	bTypes := strings.Split(rTypes.String(), ",")
```
ixdy, line 88: I think you don't need this `Split` - you can just iterate over rTypes directly (internally it's a slice)

ixdy (approved):
/hold

@detiber did you have any additional comments?

PR #46, opened 2020-10-29
GCP Janitor also cleans up pubsub resources

chizhg:
Not sure if we can add arbitrary GCP resources for Janitor to clean up, but the use case is in some of the Knative integration test flows, they create `pubsub` resources which are possible to be leaked. Currently we are running the `gcloud pubsub delete` commands before the workflow actually starts, but the best way seems to be asking Janitor to clean them up.

I'm setting `tolerate=True` here, then it won't report as failure if the cleanup command fails(?)

### DIFF
--- a/cmd/janitor/gcp_janitor.py
+++ b/cmd/janitor/gcp_janitor.py
@@ -76,6 +76,14 @@ RESOURCES_BY_API = {
         Resource('', 'logging', 'sinks', None, None, None, False, False, ['_Default', '_Required']),
     ],
 
+    # pubsub resources
+    'pubsub.googleapis.com': [
+        Resource('', 'pubsub', 'subscriptions', None, None, None, False, True, None),
+        Resource('', 'pubsub', 'topics', None, None, None, False, True,
+                 ['container-analysis-notes-v1', 'container-analysis-notes-v1beta1',
+                  'container-analysis-occurrences-v1', 'container-analysis-occurrences-v1beta1'])
+    ],
+
     # GKE hub memberships
     'gkehub.googleapis.com': [
         Resource('', 'container', 'hub', 'memberships', None, None, False, False, None),


ixdy:
Do you expect cleanup of pubsub resources to fail frequently?

Setting `tolerate=True` means that projects will be deemed clean (and returned to `free` state) even if we fail to clean up pubsub resources.

cmd/janitor/gcp_janitor.py
```diff
@@ -76,6 +76,12 @@
         Resource('', 'logging', 'sinks', None, None, None, False, False, ['_Default', '_Required']),
     ],
 
+    # pubsub resources
+    'pubsub.googleapis.com': [
+        Resource('', 'pubsub', 'subscriptions', None, None, None, True, True, None),
+        Resource('', 'pubsub', 'topics', None, None, None, True, True, None)
```
ixdy, line 82: are there certain topics we shouldn't delete? looking at a random sampling of projects, I see some with topics like
- container-analysis-notes-v1
- container-analysis-notes-v1beta1
- container-analysis-occurrences-v1
- container-analysis-occurrences-v1beta1

I don't really understand why there are container analysis pubsub topics in boskos test projects, but I also don't understand if there will be any issues if we delete them.

chizhg: Hmm I see, I did some search and found https://cloud.google.com/container-registry/docs/pub-sub-notifications#create_pubsub_topics, it looks they are automatically created after enabling the Container Analysis API, so we cannot simply ask Janitor to clean them up.

Also it looks less ideal to me to add them as `preserved_names`, since we don't know if these are all the pubsub resources we need to preserve, and we don't know if someday their names will be changed.

Considering it from another angle, would it make sense to make Janitor more customizable? Since different projects might have different requirements on what resources they want to clean up / preserve, we can by default get Janitor cleaning up some heavy-weight resources (like GKE and GCE), but also allowing users to configure additional resources they want to clean up?

chizhg:
> Do you expect cleanup of pubsub resources to fail frequently?
> 
> Setting `tolerate=True` means that projects will be deemed clean (and returned to `free` state) even if we fail to clean up pubsub resources.

I think setting `tolerate=True` is required for a non-breaking change. Since users could be using a service account that does not have pubsub admin roles then they will get errors in cleaning up pubsub resources for sure.

ixdy:
> I think setting `tolerate=True` is required for a non-breaking change. Since users could be using a service account that does not have pubsub admin roles then they will get errors in cleaning up pubsub resources for sure.

Do you need a pubsub admin role to delete pubsub resources? Or is project editor enough? (I think in many cases the Boskos janitor has at least project editor permissions.)

cmd/janitor/gcp_janitor.py
chizhg: I'm tending to leave these topics uncleaned for safety reasons, so I added them to `preserved_names`.

chizhg:
> > I think setting `tolerate=True` is required for a non-breaking change. Since users could be using a service account that does not have pubsub admin roles then they will get errors in cleaning up pubsub resources for sure.
> 
> Do you need a pubsub admin role to delete pubsub resources? Or is project editor enough? (I think in many cases the Boskos janitor has at least project editor permissions.)

I just checked the `editor` role, it does contain permissions to delete pubsub resources, so it should be fine to set `tolerate=False` here.

ixdy (approved):

PR #51, opened 2020-11-19
Mark and sweep elastic file systems

chrishenzie:
This PR adds support for cleaning up [Elastic File Systems](https://aws.amazon.com/efs/).

In order to delete file systems we must first delete all associated mount targets and wait for them to be deleted.

This `MarkAndSweep()` implementation is a little long; let me know if you'd prefer this broken into smaller functions.

@msau42

### DIFF
--- a/aws-janitor/resources/efs.go
+++ b/aws-janitor/resources/efs.go
@@ -1,0 +1,218 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package resources
+
+import (
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/request"
+	"github.com/aws/aws-sdk-go/service/efs"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	maxRetries   = 120
+	pollInterval = time.Second
+)
+
+// ElasticFileSystems: https://docs.aws.amazon.com/sdk-for-go/api/service/efs/#EFS.DescribeFileSystems
+
+type ElasticFileSystems struct{}
+
+func (ElasticFileSystems) MarkAndSweep(opts Options, set *Set) error {
+	logger := logrus.WithField("options", opts)
+	svc := efs.New(opts.Session, aws.NewConfig().WithRegion(opts.Region))
+
+	// Paged calls, defer deletion until we have the whole list.
+	var (
+		fileSystemsToDelete  []*elasticFileSystem
+		mountTargetsToDelete []*mountTarget
+	)
+
+	// Mark and sweep file systems for deletion.
+	fsPageFunc := func(page *efs.DescribeFileSystemsOutput, _ bool) bool {
+		for _, fs := range page.FileSystems {
+			f := &elasticFileSystem{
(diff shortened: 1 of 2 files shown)


msau42:
/ok-to-test
cc @wongma7

msau42:
Do you know if snapshots are already cleaned up too? If not, may be worth adding in a followup

aws-janitor/resources/efs.go
```diff
@@ -0,0 +1,210 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
ixdy, line 2: nit: this could be 2020 now

aws-janitor/resources/efs.go
```diff
@@ -0,0 +1,210 @@
...
+				}
+			}
+		}
+		return true
+	}
+	if err := svc.DescribeFileSystemsPages(&efs.DescribeFileSystemsInput{}, fsPageFunc); err != nil {
+		return err
+	}
+
+	// Mark and sweep mount targets for deletion.
```
ixdy, line 67: it looks like this comment is a little misleading - it appears that this deletes all mount targets for a filesystem, rather than marking + sweeping based on age.

aws-janitor/resources/efs.go
```diff
@@ -0,0 +1,210 @@
...
+const (
+	maxRetries   = 120
+	pollInterval = time.Second
+)
+
+// ElasticFileSystems: https://docs.aws.amazon.com/sdk-for-go/api/service/efs/#EFS.DescribeFileSystems
+
+type ElasticFileSystems struct{}
+
+func (ElasticFileSystems) MarkAndSweep(opts Options, set *Set) error {
```
ixdy, line 38: per the lint error, it might be worth breaking this function up a bit - maybe the handling of mount targets should be a separate function?

(that said, I probably need to tune that linter a bit.)

aws-janitor/resources/efs.go
```diff
@@ -0,0 +1,210 @@
...
+	for _, mt := range mountTargetsToDelete {
+		deleteInput := &efs.DeleteMountTargetInput{
+			MountTargetId: aws.String(mt.ID),
+		}
+		if _, err := svc.DeleteMountTarget(deleteInput); err != nil {
+			logger.Warningf("%s: delete failed: %v", mt.ID, err)
+		}
+	}
+
+	// Wait for mount targets to be deleted.
```
ixdy, line 98: it might be worth adding a `logger.Debug` call before entering this loop.

also, do you have any stats on how long this usually takes?

ixdy (commented):
basically LGTM, just have a few comments. thanks for this PR!

aws-janitor/resources/efs.go
chrishenzie: It's reasonable; I felt this function was getting too large as I was writing it, but I didn't like the alternative of passing everything as arguments to helper functions.

It would be nice to make things like the EFS client and logger struct fields so we don't have to pass everything around as args, but that requires a bit more refactoring (making constructors for all of these types). If you have any other ideas for working around this let me know.

chrishenzie: I split this into a separate function. I had thought about doing this for each stage of `MarkAndSweep()`. I could proceed with this, but I was reluctant because the "style" would stray from what we're doing for other resource types.

aws-janitor/resources/efs.go
chrishenzie: No exact numbers, but usually takes under a minute.

chrishenzie: Just tested, only took 10-15 seconds to wait for it to be deleted.

chrishenzie:
Tested locally by commenting out all items in `RegionalTypeList` except for `ElasticFileSystems`, and all items in `GlobalTypeList` and ran:
```
# Listed all EFS file systems (each had 4 mount targets)
go run cmd/aws-janitor/main.go -all -region us-west-2 -dry-run true

# Saw it delete the mount targets and file systems
go run cmd/aws-janitor/main.go -all -region us-west-2
```

chrishenzie:
Opened https://github.com/kubernetes-sigs/boskos/pull/52 to also clean up volume snapshots since these are also created through E2E tests.

ixdy (approved):

PR #52, opened 2020-11-21
Mark and sweep EBS volume snapshots

chrishenzie:
This PR adds support for cleaning up [EBS volume snapshots](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSSnapshots.html).

@msau42 @ixdy

### DIFF
--- a/aws-janitor/resources/snapshots.go
+++ b/aws-janitor/resources/snapshots.go
@@ -1,0 +1,107 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


chrishenzie:
This actually needs some tweaking. It looks like this also returns publicly available snapshots. Working on filtering this down to just account-owned snapshots.

aws-janitor/resources/snapshots.go
```diff
@@ -0,0 +1,107 @@
...
+	return set, errors.Wrapf(err, "couldn't describe snapshots for %q in %q", opts.Account, opts.Region)
+}
+
+type snapshot struct {
+	// The current client library does not provide ARNs for snapshots.
+	ID string
+}
+
+func (s snapshot) ARN() string {
+	return s.ID
```
ixdy, line 102: it occurs to me that this might be buggy (for how it's used in `set.Mark`) if there are other resource types with the same ID, or the same ID for this resource type across multiple regions. I'm not sure how likely this is, though, and this problem isn't unique to this resource, so you don't really need to worry about it in this PR.

(mostly a note to myself that I should investigate and clean this up.)

ixdy (approved):

aws-janitor/resources/snapshots.go
ixdy: #55

PR #54, opened 2020-11-23
Fix configsync for static -> dynamic -> static

alvaroaleman:
First commit: Make config sync a controller
* We need to retry on errors
* We need to react to changes in resources, specifically Add/Delete or Update where a resource gets set into `Tombstome` state

Second commit: Fix config sync when a config is changed from static -> dynamic -> static
This currently causes an `IsAlreadyExists` error because the calculation for the static resource creation ignores all resources that have a corresponding dynamic config. This blocks any further progress.
-> Fixed by including all resources that have a static config in the static resource add/delete calculation
-> This made the dynamic resource reconciliation instantly delete them. Fixed by making the dynamic resource reconciliation ignore all resources that have a static config

Note: While I added a unit testcase for this specific problem, I verified the pull by deploying boskos and changing its config from (for the same resource) static -> dynamic -> static -> dyanamic -> static in order to make sure that all of reconciliation and triggering work correctly and that there are no errors.

/assign @ixdy

### DIFF
--- a/cmd/boskos/boskos.go
+++ b/cmd/boskos/boskos.go
@@ -201,7 +201,7 @@ func addConfigSyncReconcilerToManager(mgr manager.Manager, configSync func() err
 		return fmt.Errorf("failed to watch boskos resources: %w", err)
 	}
 	if err := ctrl.Watch(&source.Kind{Type: &crds.DRLCObject{}}, constHandler()); err != nil {
-		return fmt.Errorf("failed to watch boskos dyanmic resources: %w", err)
+		return fmt.Errorf("failed to watch boskos dynamic resources: %w", err)
 	}
 	if err := ctrl.Watch(&source.Channel{Source: configChangeEvent}, constHandler()); err != nil {
 		return fmt.Errorf("failed to create source channel for config change event: %w", err)
--- a/cmd/boskos/boskos.go
+++ b/cmd/boskos/boskos.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"context"
 	"flag"
 	"fmt"
 	"net/http"
@@ -28,6 +29,13 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/viper"
 	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	"k8s.io/test-infra/pkg/flagutil"
 	"k8s.io/test-infra/prow/config"
@@ -36,6 +44,7 @@ import (
 	"k8s.io/test-infra/prow/logrusutil"
 	prowmetrics "k8s.io/test-infra/prow/metrics"
 	"k8s.io/test-infra/prow/pjutil"
+	"sigs.k8s.io/boskos/common"
 	"sigs.k8s.io/boskos/crds"
 	"sigs.k8s.io/boskos/handlers"
 	"sigs.k8s.io/boskos/metrics"
@@ -49,9 +58,9 @@ const (
 )
 
 var (
-	configPath                  = flag.String("config", "config.yaml", "Path to init resource file")
-	dynamicResourceUpdatePeriod = flag.Duration("dynamic-resource-update-period", defaultDynamicResourceUpdatePeriod,
-		"Period at which to update dynamic resources. Set to 0 to disable.")
+	configPath = flag.String("config", "config.yaml", "Path to init resource file")
+	_          = flag.Duration("dynamic-resource-update-period", defaultDynamicResourceUpdatePeriod,
+		"Legacy flag that does nothing but is kept for compatibility reasons")
 	requestTTL = flag.Duration("request-ttl", defaultRequestTTL, "request TTL before losing priority in the queue")
 	logLevel   = flag.String("log-level", "info", fmt.Sprintf("Log level is one of %v.", logrus.AllLevels))
 	namespace  = flag.String("namespace", corev1.NamespaceDefault, "namespace to install on")
@@ -99,12 +108,28 @@ func main() {
 	// main server with the main mux until we're ready
 	health := pjutil.NewHealth()
 
-	client, err := kubeClientOptions.CacheBackedClient(*namespace, &crds.ResourceObject{}, &crds.DRLCObject{})
+	cfg, err := kubeClientOptions.Cfg()
 	if err != nil {
-		logrus.WithError(err).Fatal("unable to get client")
+		logrus.WithError(err).Fatal("Failed to get kubeconfig")
 	}
+	cfg.QPS = 100
+	cfg.Burst = 200
+	mgr, err := manager.New(cfg, manager.Options{
+		LeaderElection:     false,
+		Namespace:          *namespace,
+		MetricsBindAddress: "0",
+	})
+	if err != nil {
+		logrus.WithError(err).Fatal("Failed to construct mgr.")
+	}
+	interrupts.Run(func(ctx context.Context) {
+		if err := mgr.Start(ctx.Done()); err != nil {
+			logrus.WithError(err).Fatal("Mgr failed.")
+		}
+		logrus.Info("Mgr finished gracefully.")
+	})
 
-	storage := ranch.NewStorage(interrupts.Context(), client, *namespace)
+	storage := ranch.NewStorage(interrupts.Context(), mgr.GetClient(), *namespace)
 
 	r, err := ranch.NewRanch(*configPath, storage, *requestTTL)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


ranch/ranch.go
```diff
@@ -91,12 +91,6 @@ func NewRanch(config string, s *Storage, ttl time.Duration) (*Ranch, error) {
 		requestMgr: NewRequestManager(ttl),
 		now:        time.Now,
 	}
-	if config != "" {
-		if err := newRanch.SyncConfig(config); err != nil {
```
alvaroaleman, line 95: This races with the cache startup and will fail if the cache wasn't started in time

cmd/boskos/boskos.go
```diff
@@ -144,3 +172,69 @@ func main() {
...
+	return predicate.Funcs{
+		CreateFunc: func(_ event.CreateEvent) bool { return true },
+		DeleteFunc: func(_ event.DeleteEvent) bool { return true },
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			resource, ok := e.ObjectNew.(*crds.ResourceObject)
+			if !ok {
+				panic(fmt.Sprintf("BUG: expected *crds.ResourceObject, got %T", e.ObjectNew))
+			}
+
+			return resource.Status.State == common.Tombstone
```
alvaroaleman, line 236: This is needed because boskos is the entity that deletes Tombstoned resources: https://github.com/kubernetes-sigs/boskos/blob/eea29b1a07663e131ea8eb5c8f8b6537e36ad8ee/ranch/storage.go#L291

The process is:
* Boskos sets to ToBeDeleted
* Cleaner cleans and sets to Tombstone
* Boskos deletes

To me it seems that the cleaner should just delete them. The only additional thing we do here is to check that they have no owner, but they will always have no owner because the cleaner always does both clean, then release: https://github.com/kubernetes-sigs/boskos/blob/eea29b1a07663e131ea8eb5c8f8b6537e36ad8ee/cleaner/cleaner.go#L88-L90

Are there more reasons or can we (in a follow-up) make the cleaner just delete the resources?

cmd/boskos/boskos.go
```diff
@@ -144,3 +172,69 @@ func main() {
 	// signal to the world that we're ready
 	health.ServeReady()
 }
+
+type configSyncReconciler struct {
```
ixdy, line 176: would it make sense to move the rest of this out of `cmd/boskos/boskos.go`?

cmd/boskos/boskos.go
ixdy: cc @sebastienvas 

This seems reasonable to me. There may have been some reason this was separated that got removed in subsequent refactorings.

ranch/ranch.go
```diff
@@ -410,25 +404,6 @@ func (r *Ranch) SyncConfig(configPath string) error {
 	return r.Storage.SyncResources(config)
 }
 
-// StartDynamicResourceUpdater starts a goroutine which periodically
-// updates all dynamic resources.
-func (r *Ranch) StartDynamicResourceUpdater(updatePeriod time.Duration) {
```
ixdy, line 415: Am I understanding correctly that the reason we no longer need this is because we'll basically be syncing all resources anytime one of the dynamic resources is created or deleted?

(This code was added in https://github.com/kubernetes/test-infra/issues/15906, and I'm trying to convince myself that it still does the appropriate thing. I guess the unit tests help enforce that.)

cmd/boskos/boskos.go
```diff
@@ -144,3 +172,69 @@ func main() {
...
+
+func constHandler() handler.EventHandler {
+	return &handler.EnqueueRequestsFromMapFunc{
+		ToRequests: handler.ToRequestsFunc(func(handler.MapObject) []reconcile.Request {
+			return []reconcile.Request{{}}
+		}),
+	}
+}
+
+// resourceUpdatePredicate prevents the config reconciler from reacting to resource update events
```
ixdy, line 224: it doesn't need to happen in this PR, but it seems like we might be able to tweak this to fix https://github.com/kubernetes-sigs/boskos/issues/20 as well.

ranch/storage.go
```diff
@@ -140,7 +141,21 @@ func (s *Storage) DeleteDynamicResourceLifeCycle(name string) error {
 			Namespace: s.namespace,
 		},
 	}
-	return s.client.Delete(s.ctx, o)
+	if err := s.client.Delete(s.ctx, o); err != nil {
+		return err
+	}
+	if err := wait.Poll(100*time.Millisecond, 5*time.Second, func() (bool, error) {
```
ixdy, line 148: why do we need to do this? and where did the value of 5s come from?

ranch/storage.go
```diff
@@ -314,7 +329,11 @@ func (s *Storage) updateDynamicResources(lifecycle *crds.DRLCObject, resources [
 			toDelete = append(toDelete, res)
 		}
 	}
-	logrus.Infof("DRLC type %s: adding %+v, deleting %+v", lifecycle.Name, toAdd, toDelete)
+	logrus.WithFields(logrus.Fields{
+		"type":               lifecycle.Name,
+		"add_items_count":    len(toAdd),
+		"delete_items_count": len(toDelete),
+	}).Info("Updating DLRCs")
```
ixdy: nit: DRLCs, not DLRCs :)

ranch/storage.go
```diff
@@ -418,11 +423,25 @@ func (s *Storage) syncDynamicResourceLifeCycles(newDRLCByType, existingDRLCByTyp
 			if !reflect.DeepEqual(existingDRLC, newDRLC) {
 				dRLCToUpdate = append(dRLCToUpdate, newDRLC)
 			}
-		} else {
+		} else if existingDRLC.Spec.MinCount != 0 || existingDRLC.Spec.MaxCount != 0 {
 			// Mark for deletion of all associated dynamic resources.
 			existingDRLC.Spec.MinCount = 0
 			existingDRLC.Spec.MaxCount = 0
-			dRLCToUpdate = append(dRLCToUpdate, existingDRLC)
+			if err := s.client.Update(s.ctx, existingDRLC.DeepCopyObject()); err != nil {
```
ixdy, line 430: Why is it necessary to do this update in the loop?

(Partially I'm wondering why we treat this as a special update, rather than just passing the DRLC to deleteDynamicResourceLifecycles, though I guess we don't have the list of static resources here. Hm.)

I guess it feels like this code is a bit messy and needs some cleaning up, though that does predate this PR. Thoughts?

ixdy (commented):
Thanks for this! Generally LG, just a few questions/comments.

ranch/storage.go
```diff
@@ -399,11 +423,25 @@ func (s *Storage) syncDynamicResourceLifeCycles(newDRLCByType, existingDRLCByTyp
...
+				errs = append(errs, err)
+				continue
+			}
+			// Make sure we have this change in our cache
+			name := ctrlruntimeclient.ObjectKey{Namespace: existingDRLC.Namespace, Name: existingDRLC.Name}
+			if err := wait.Poll(100*time.Millisecond, 5*time.Second, func() (bool, error) {
+				dlrc := &crds.DRLCObject{}
```
ixdy, line 437: s/dlrc/drlc/g :)

(it's the worst acronym)

ranch/ranch.go
alvaroaleman: > Am I understanding correctly that the reason we no longer need this is because we'll basically be syncing all resources anytime one of the dynamic resources is created or deleted?

Yes, anytime any of the dynamic resources changes in any way, the config changes or resources get set to tombstone state

ranch/storage.go
alvaroaleman: The client is cache-backed, which means that it takes a moment for our own changes to be visible for us. This is important here, because otherwise we will a moment later get the drlc, not see this change, exclude the now-static resources from the number of static resources created for this dlrc, note that we have too little, create them and then a moment later attempt do delete all of them.

ranch/storage.go
alvaroaleman: I actually removed that log completely because its redundant with the log that logs each create/delete individually

ranch/storage.go
alvaroaleman: it needs to be in the loop because only for this change we must make sure to again have it in the cache before proceeding.

ranch/storage.go
alvaroaleman: The 100ms and 5s are relatively arbitrary, under normal circumstances the cache will lag between 200ms to 2s behind.

cmd/boskos/boskos.go
alvaroaleman: where to, a new package or just a different file?

cmd/boskos/boskos.go
alvaroaleman: Added that

ranch/storage.go
alvaroaleman: > I guess it feels like this code is a bit messy and needs some cleaning up, though that does predate this PR. Thoughts?

Certainly, for example the approach here of constructing a slice to then pass that to another func that serially updates all elements in the slice is just needlessly complex and adds a couple of allocations, just directly updating here would be simpler. But The change already has an about 400 line diff, so I wanted to not make that worse

cmd/boskos/boskos.go
ixdy: I was thinking "new package" but now I'm not sure it's really that necessary.

cmd/boskos/boskos.go
```diff
@@ -144,3 +172,72 @@ func main() {
...
+	})
+	if err != nil {
+		return fmt.Errorf("failed to construct controller: %w", err)
+	}
+
+	if err := ctrl.Watch(&source.Kind{Type: &crds.ResourceObject{}}, constHandler(), resourceUpdatePredicate()); err != nil {
+		return fmt.Errorf("failed to watch boskos resources: %w", err)
+	}
+	if err := ctrl.Watch(&source.Kind{Type: &crds.DRLCObject{}}, constHandler()); err != nil {
+		return fmt.Errorf("failed to watch boskos dyanmic resources: %w", err)
```
ixdy, line 204: nit: dynamic

ranch/storage.go
```diff
@@ -482,21 +544,21 @@ func (s *Storage) persistDynamicResourceLifeCycles(dRLCToUpdate, dRLCToAdd, dRLC
...
-
-	for _, dRLC := range dRLCToUpdate {
-		logrus.Infof("Updating resource type life cycle %s", dRLC.Name)
-		if _, err := s.UpdateDynamicResourceLifeCycle(&dRLC); err != nil {
-			errs = append(errs, err)
+			if err := s.client.Update(s.ctx, &dRLCToDelelete[idx]); err != nil {
+				errs = append(errs, err)
+				continue
+			}
+			// Make sure we have this change in our cache
```
ixdy, line 551: should we just put this logic in `UpdateDynamicResourceLifeCycle`, and then reduce some of the code-duplication? (I guess figuring out the right condition to check is tricky...)

couldn't we run into similar issues where an update to a DRLC (e.g. changing the mincount/maxcount to something nonzero) results in us creating and then immediately deleting a dynamic resource?

I guess it might increase the latency for syncs, but hopefully not by much?

alvaroaleman: Done (We have this logic twice, once in syncDynamicResourceLifeCycles and once in UpdateAllDynamicResources. We should probably unexport the latter and remove those bits there, as they are redundant now)

ixdy: can you replace the duplicated logic in this function now with a call to `UpdateDynamicResourceLifeCycle`? i.e. replace the `s.client.Update` call (after setting MinCount and MaxCount to 0) with `UpdateDynamicResourceLifeCycle(&dRLCToDelelete[idx])`?

(also I just noticed that typo on `Delete` in `dRLCToDelelete`)

alvaroaleman: Done and fixed the typo

alvaroaleman:
@ixdy any chance you can have another look?

ixdy (approved):
Apologies for the long review latency on this one. LGTM!

PR #59, opened 2020-12-11
aws-janitor: use creation timestamp when available

ixdy:
Many resources don't have a readily accessible creation timestamp in their metadata, but at least some of them do, so we may as well make use of it when available.

/assign @detiber 
cc @chemikadze

### DIFF
--- a/aws-janitor/resources/instance.go
+++ b/aws-janitor/resources/instance.go
@@ -54,8 +54,8 @@ func (Instances) MarkAndSweep(opts Options, set *Set) error {
 					Region:     opts.Region,
 					InstanceID: *inst.InstanceId,
 				}
-
-				if set.Mark(i) {
+				// Instances don't have a creation date, but launch time is better than nothing.
+				if set.Mark(i, inst.LaunchTime) {
 					logger.Warningf("%s: deleting %T: %s", i.ARN(), inst, i.InstanceID)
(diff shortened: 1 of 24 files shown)


aws-janitor/resources/instance.go
```diff
@@ -54,8 +54,8 @@ func (Instances) MarkAndSweep(opts Options, set *Set) error {
 					Region:     opts.Region,
 					InstanceID: *inst.InstanceId,
 				}
-
-				if set.Mark(i) {
+				// Instances don't have a creation date, but launch time is better than nothing.
```
chemikadze, line 57: Is that practically same thing? (I believe instances use LaunchXxx API call instead of CreateXxx)

chemikadze (approved):

aws-janitor/resources/instance.go
ixdy: I think for our purposes, they'll generally be the same thing, yes. My reference is [this SO thread](https://stackoverflow.com/questions/18916135/determining-amazon-ec2-instance-creation-date-time), which notes that instances can potentially be restarted (without being recreated), which would reset this timestamp, but I doubt most tests will be doing that.

richardcase (approved):
Seems ok to me, so:  

/lgtm

PR #60, opened 2020-12-12
aws-janitor: add support for EKS clusters

ixdy:
/hold
(since this is dependent on #59)

/assign @detiber 
cc @chemikadze

### DIFF
--- a/aws-janitor/resources/eks.go
+++ b/aws-janitor/resources/eks.go
@@ -1,0 +1,151 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package resources
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/eks"
+	"github.com/sirupsen/logrus"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
+)
+
+type EKS struct{}
+
+func (e EKS) MarkAndSweep(opts Options, set *Set) error {
+	logger := logrus.WithField("options", opts)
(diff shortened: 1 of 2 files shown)


aws-janitor/resources/eks.go
```diff
@@ -0,0 +1,98 @@
...
+		if !opts.DryRun {
+			toDelete = append(toDelete, cluster)
+		}
+	})
+	if err != nil {
+		return err
+	}
+
+	for _, cluster := range toDelete {
+		if _, err := svc.DeleteCluster(&eks.DeleteClusterInput{Name: aws.String(cluster.name)}); err != nil {
```
richardcase, line 43: I think we may need to check to see if there are any managed node groups attached to the cluster and delete those first.

richardcase: So perhaps `ListNodegroups` and then for each fo `DeleteNodegroup`

randomvariable: Do you want to do this as follow up, or amend this PR?

richardcase: For it to clean EKS clusters created using the `AWSManagedMachinePool` this will be required. 

@ixdy - did you want me to add the changes for deleting node groups or did you want to do it?

ixdy: I'll add this functionality and update this PR later this week - I've been on vacation the last few weeks. Thanks for the review!

randomvariable: No probs @ixdy. We were on hols too, and was going to pick this up this week and saw you had already started :)

ixdy: OK, this PR has been updated with your suggestion. Like most of the rest of the janitor, if the deletes fail, it just logs a warning and continues. I'm not sure what would happen if the nodegroup deletion fails but the cluster deletion succeeds. (Will the nodegroup get leaked in that case?) Or will it just leak the EC2 VMs, which should get handled via later stages?

ixdy: After reading the [docs](https://docs.aws.amazon.com/eks/latest/userguide/delete-cluster.html), I decided to make this handle fargate profiles too.

Additionally, I made it skip trying to delete EKS clusters if there were any failures in deleting nodegroups or fargate profiles, assuming that a later run might succeed.

ixdy:
/hold cancel

as previous PR has merged

aws-janitor/resources/eks.go
richardcase: Thats great @ixdy.

If the nodegroup deletion fails then the cluster delete will fail. Good call on adding the fargate profiles delete as well.

richardcase (approved):
Seems good to me.

/lgtm

Issue #55, opened 2020-11-23
aws-janitor: ensure IDs are unique across resources and regions for set.Mark


ixdy:
Some AWS resources don't have an ARN, so we're currently just using the ID of the resource, which may not necessarily be globally unique for the same resource and/or across resource types.

For several resource types, we generate a fake ARN which includes the resource type and region. We should ensure all resource types are doing something similar.

/kind cleanup

cpanato:
hello @ixdy 
I would like to contribute to this, can you point to me one example where this is generated for a resource?
thanks!

ixdy:
Yep, that's a perfect example. Thanks!

/assign @cpanato

cpanato:
@ixdy thanks! i will work on that

PR #67, opened 2021-01-20
aws-janitor: normalize generated arn ids for resources

cpanato:
Normalize the ARN for resources that do not have.

/assign @ixdy 

please let me know if that makes sense or what you think, thanks for your review in advance :)

Fixes: https://github.com/kubernetes-sigs/boskos/issues/55

### DIFF
--- a/aws-janitor/resources/asg.go
+++ b/aws-janitor/resources/asg.go
@@ -37,9 +37,12 @@ func (AutoScalingGroups) MarkAndSweep(opts Options, set *Set) error {
 
 	pageFunc := func(page *autoscaling.DescribeAutoScalingGroupsOutput, _ bool) bool {
 		for _, asg := range page.AutoScalingGroups {
-			a := &autoScalingGroup{ID: *asg.AutoScalingGroupARN, Name: *asg.AutoScalingGroupName}
+			a := &autoScalingGroup{
+				arn:  aws.StringValue(asg.AutoScalingGroupARN),
+				name: aws.StringValue(asg.AutoScalingGroupName),
+			}
 			if set.Mark(a, asg.CreatedTime) {
-				logger.Warningf("%s: deleting %T: %s", a.ARN(), asg, a.Name)
+				logger.Warningf("%s: deleting %T: %s", a.ARN(), asg, a.name)
 				if !opts.DryRun {
 					toDelete = append(toDelete, a)
 				}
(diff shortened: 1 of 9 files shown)


aws-janitor/resources/asg.go
```diff
@@ -103,12 +106,14 @@ func (AutoScalingGroups) ListAll(opts Options) (*Set, error) {
...
-	ID   string
-	Name string
+	Account string
+	Region  string
+	ID      string
+	Name    string
 }
 
 func (asg autoScalingGroup) ARN() string {
-	return asg.ID
```
ixdy, line 111: I think this one is already using the ARN (see L41), it's just named improperly in the struct.

aws-janitor/resources/efs.go
```diff
@@ -123,12 +126,13 @@ func (ElasticFileSystems) ListAll(opts Options) (*Set, error) {
...
 
 type elasticFileSystem struct {
-	ID string
+	Account string
+	Region  string
+	ID      string
 }
 
 func (efs elasticFileSystem) ARN() string {
-	// The current client library does not provide ARNs for file systems.
```
ixdy, line 130: I think this comment is actually false? (or maybe it was added at some point?) 

we should be able to use the ARN provided in the result: https://docs.aws.amazon.com/sdk-for-go/api/service/efs/#FileSystemDescription

aws-janitor/resources/cloud_formation_stacks.go
```diff
@@ -87,12 +96,14 @@ func (CloudFormationStacks) ListAll(opts Options) (*Set, error) {
...
-	name string
+	account string
+	region  string
+	id      string
+	name    string
 }
 
 func (p cloudFormationStack) ARN() string {
-	return p.id
+	return fmt.Sprintf("arn:aws:cloud-formation:%s:%s:stack/%s", p.region, p.account, p.id)
```
ixdy, line 106: nit: it seems like most of the real ARNs don't use dashes in the service name (i.e. make this `cloudformation`, not `cloud-formation`)

aws-janitor/resources/elb.go
```diff
@@ -96,7 +97,7 @@ type classicLoadBalancer struct {
 }
 
 func (lb classicLoadBalancer) ARN() string {
-	return "fakearn:elb:" + lb.region + ":" + lb.account + ":" + lb.dnsName
+	return fmt.Sprintf("arn:aws:ec2:%s:%s:classic-elb/%s", lb.region, lb.account, lb.dnsName)
```
ixdy, line 100: nit: service probably should be something like `elb`, not `ec2`

ixdy (commented):
thanks! just a few comments.

aws-janitor/resources/efs.go
cpanato: to use that, i've bumped the `github.com/aws/aws-sdk-go` dependency

aws-janitor/resources/asg.go
cpanato: true! thanks for pointing out

thanks for your review @ixdy 
PTAL

ixdy (approved):
Sorry for the long review cycle on this. Thanks for the PR, I do appreciate it!

PR #68, opened 2021-01-20
hack: drop hack go modules

cpanato:
Some projects in the cluster-api we already removed the hack/tools.go and go modules for the hack and move that to the makefile, I think that makes a bit cleaner

if you don't agree you can close the PR 😄 

updated golang-ci and gotestsum to the latest releases

/assign @ alvaroaleman @ixdy


ixdy:
Do you know why those projects switched? I'm not completely opposed, but this seems like more infrastructure to maintain and doesn't bring obvious benefits that I can see.

cpanato:
this is a good description: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/619#issue-418299662

Makefile
```diff
@@ -19,11 +19,17 @@ OUTPUT_DIR ?= _output
 
 TOOLS_DIR := hack/tools
 BIN_DIR := bin
-TOOLS_BIN_DIR := $(TOOLS_DIR)/$(BIN_DIR)
+TOOLS_BIN_DIR := $(abspath $(TOOLS_DIR)/$(BIN_DIR))
 OUTPUT_BIN_DIR := $(OUTPUT_DIR)/$(BIN_DIR)
+GO_INSTALL = ./scripts/go_install.sh
```
ixdy, line 24: nit: maybe put this script in `hack/` instead of `scripts/`?

I know it doesn't entirely match the go standard project layout, but many k8s projects seem to use `hack` instead of `scripts`, and we already have some scripts in `hack`.

scripts/go_install.sh
```diff
@@ -0,0 +1,56 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+if [ -z "${1}" ]; then
```
ixdy, line 20: nit: prefer the bash builtins `[[ ]]` over `[ ]`

scripts/go_install.sh
```diff
@@ -0,0 +1,56 @@
...
+
+rm "${GOBIN}/${2}"* || true
+
+cd "${tmp_dir}"
+
+# create a new module in the tmp directory
+go mod init fake/mod
+
+# install the golang module specified as the first argument
+go get -tags tools "${1}@${3}"
```
ixdy, line 54: minor suggestion: mapping `$1`, `$2`, and `$3` to better-named variables will make this easier to read (and likely less error-prone)

ixdy (commented):
Thanks for the reference. tl;dr for how I understood it: using the `hack/tools` trick means that all tools will have to use the same versions of dependencies, which can cause problems (though it may not be for boskos, yet?).

This change overall LGTM, though I have a few minor nits/suggestions.

scripts/go_install.sh
cpanato: done, thanks for your review!
PTAL

hack/go_install.sh
```diff
@@ -0,0 +1,59 @@
...
+  exit 1
+fi
+
+tmp_dir=$(mktemp -d -t goinstall_XXXXXXXXXX)
+function clean {
+  rm -rf "${tmp_dir}"
+}
+trap clean EXIT
+
+rm "${GOBIN}/${2}"* || true
```
ixdy, line 49: nit: replace `${2}` with `${binary_name}` (also on L58)

cpanato: Oh, my bad :/

cpanato: done. PTAL

ixdy (approved):

PR #69, opened 2021-02-03
aws-janitor: Cleanup SQS queues and EventBridge rules

sedefsavas:
This PR adds aws-janitor to clean up SQS queues and  EventBridge rules.

### DIFF
--- a/aws-janitor/resources/sqs.go
+++ b/aws-janitor/resources/sqs.go
@@ -1,0 +1,178 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package resources
+
+import (
+	"fmt"
+	"strconv"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/eventbridge"
+	"github.com/aws/aws-sdk-go/service/sqs"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+// SQS queues: https://docs.aws.amazon.com/sdk-for-go/api/service/sqs
+
+type SQSQueues struct{}
+
+func (SQSQueues) MarkAndSweep(opts Options, set *Set) error {
+	logger := logrus.WithField("options", opts)
+	svc := sqs.New(opts.Session, aws.NewConfig().WithRegion(opts.Region))
+
+	input := &sqs.ListQueuesInput{}
+
+	var toDelete []*sqsQueue // Paged call, defer deletion until we have the whole list.
+
+	pageFunc := func(page *sqs.ListQueuesOutput, _ bool) bool {
+		for _, url := range page.QueueUrls {
+			attrInput := &sqs.GetQueueAttributesInput{
+				AttributeNames: []*string{aws.String(sqs.QueueAttributeNameAll)},
+				QueueUrl:       url,
+			}
+			attr, err := svc.GetQueueAttributes(attrInput)
+			if err != nil {
+				return false
+			}
+
+			q := &sqsQueue{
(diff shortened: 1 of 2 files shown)


sedefsavas:
/hold 
while testing

sedefsavas:
Combined SQS queue and rule cleanup into one logic to filter rules that selected SQS queues as target.
Now, ready to go.

aws-janitor/resources/sqs.go
```diff
@@ -0,0 +1,178 @@
...
+			// Only delete rules that uses SQS queue as target. There are default rules that should not be deleted.
+			rules, err := svcRules.ListRuleNamesByTarget(&eventbridge.ListRuleNamesByTargetInput{
+				TargetArn: attr.Attributes[sqs.QueueAttributeNameQueueArn],
+			})
+			if err != nil {
+				logger.Warningf("listing rules by target failed: %s", err.Error())
+			}
+
+			for _, rule := range rules.RuleNames {
+				deleteEventBridgeRule(rule, svcRules, logger)
```
ixdy, line 72: it looks like this will remove targets and delete rules even when dryrun is set - is that correct?

aws-janitor/resources/sqs.go
```diff
@@ -0,0 +1,178 @@
...
+			}
+
+			unitTimestamp, _ := strconv.ParseInt(*attr.Attributes[sqs.QueueAttributeNameCreatedTimestamp], 10, 64)
```
ixdy, line 75: I think this might want to be `aws.StringValue` instead of simply dereferencing, since if `QueueAttributeNameCreatedTimestamp` is missing for some reason, this would be a nil dereference.

aws-janitor/resources/sqs.go
```diff
@@ -0,0 +1,178 @@
...
+				return false
+			}
+
+			q := &sqsQueue{
+				Account:  opts.Account,
+				Region:   opts.Region,
+				Name:     *attr.Attributes[sqs.QueueAttributeNameQueueArn],
+				QueueURL: *url,
+			}
+			svcRules := eventbridge.New(opts.Session, aws.NewConfig().WithRegion(opts.Region))
```
ixdy, line 61: do we need to create this service every iteration of the loop? or should it be created outside the loop (or even outside the page function)?

aws-janitor/resources/sqs.go
```diff
@@ -0,0 +1,178 @@
...
+
+			// Only delete rules that uses SQS queue as target. There are default rules that should not be deleted.
+			rules, err := svcRules.ListRuleNamesByTarget(&eventbridge.ListRuleNamesByTargetInput{
```
ixdy, line 64: I think this only returns up to 100 rules. do we expect this to be an issue?

aws-janitor/resources/sqs.go
```diff
@@ -0,0 +1,178 @@
...
+func deleteEventBridgeRule(rule *string, svcRules *eventbridge.EventBridge, logger *logrus.Entry) {
+	// Before removing a rule, all the target must be removed from the rule.
+	// For removing the targets, target ids must be provided.
+	targets, err := svcRules.ListTargetsByRule(&eventbridge.ListTargetsByRuleInput{
+		Rule: rule,
+	})
+	if err != nil {
+		logger.Warningf("%s: listing targets failed: %v", *rule, err)
+	}
+	targetStr := make([]*string, 0)
```
ixdy, line 113: nit: instead of size 0, you could make this `len(targets.Targets))`

aws-janitor/resources/sqs.go
```diff
@@ -0,0 +1,178 @@
...
+			logger.Warningf("%s: delete failed: %v", q.ARN(), err)
+		}
+	}
+	return nil
+}
+
+func deleteEventBridgeRule(rule *string, svcRules *eventbridge.EventBridge, logger *logrus.Entry) {
+	// Before removing a rule, all the target must be removed from the rule.
+	// For removing the targets, target ids must be provided.
+	targets, err := svcRules.ListTargetsByRule(&eventbridge.ListTargetsByRuleInput{
```
ixdy, line 107: I think this similarly lists only 100 targets at a time.

aws-janitor/resources/sqs.go
```diff
@@ -0,0 +1,178 @@
...
+			}
+			attr, err := svc.GetQueueAttributes(attrInput)
+			if err != nil {
+				return false
+			}
+
+			q := &sqsQueue{
+				Account:  opts.Account,
+				Region:   opts.Region,
+				Name:     *attr.Attributes[sqs.QueueAttributeNameQueueArn],
```
ixdy, line 58: is `Name` here being set to the full `ARN`? if so, the `ARN()` function might be returning the wrong value.

ixdy (commented):
Thanks for this! I just have a few follow-up comments.

Issue #13, opened 2020-05-29
monitor boskos cleanup timing


ixdy:
*Originally filed as https://github.com/kubernetes/test-infra/issues/14715 by @BenTheElder*

**What would you like to be added**: export and graph metrics for boskos cleanup timing

**Why is this needed**: so we can determine if this is increasing and we need to increase the janitor or fix boskos xref #14697 

Possibly this should also move to the new monitoring stack? cc @cjwagner @detiber 

/area boskos
/assign @krzyzacy 
cc @fejta @mm4tt 
/kind feature

cpanato:
this is something to add a prometheus metric for this operation? @detiber

detiber:
@cpanato I believe that to be the case, yes. That said, I haven't dug into how the existing metrics are exposed for boskos. The dashboards sit at monitoring.prow.k8s.io, though.

cpanato:
hello @ixdy the metric in question should be added in this part https://github.com/kubernetes-sigs/boskos/tree/master/cmd/cleaner ? or it is for another part of the code?
maybe the first question is, this is still needed?

ixdy:
Sorry for the delay in response. To clarify, this would be metrics added to the janitor(s), not the (unfortunately named) cleaner component.

The basic gist is just adding some Prometheus metrics to the janitors, yes, but the primary challenge is that in some deployments (such as k8s.io prow) Boskos + the janitors run in a completely separate build cluster from the prow monitoring stack, which makes collecting these metrics more challenging, since they aren't directly accessible.

In the case of k8s.io prow, to collect metrics from the core boskos service, we [expose the boskos metrics port on an external IP](https://github.com/kubernetes/test-infra/blob/be1ab1b753ca01fe554457e84b9465344bc41cb8/config/prow/cluster/boskos.yaml#L171) and then [explicitly collect from that address](https://github.com/kubernetes/test-infra/blob/be1ab1b753ca01fe554457e84b9465344bc41cb8/config/prow/cluster/monitoring/additional-scrape-configs_secret.yaml#L14). Since the janitors run as a separate container, we'd need to either expose additional IPs for each janitor (non-ideal) or set up some sort of collector for all of the boskos metrics (core and janitor) and then expose that to the prow monitoring stack. Alternately, we could collect/push these metrics to the monitoring stack. [Note: I'm probably using the wrong Prometheus terminology here.]

Figuring all of this out is the harder aspect of this issue. If this sounds interesting to you, please take it on!

[...]

PR #75, opened 2021-02-24
aws-janitor-boskos: add clean time and process time metrics 

cpanato:
...to collect the response time when cleaning and for the entire process when doing the janitor process

Make the histogram and created the exponential buckets, I've spoken in the past with `Frederic Branczyk - @brancz` and he suggests following this approach since it is very similar to what I've implemented in the past then used the same 😄 

this is part of https://github.com/kubernetes-sigs/boskos/issues/13

opening this PR for feedback and see if sounds good, and then will do it for the other janitors

/assign @ixdy

### DIFF
--- a/cmd/aws-janitor-boskos/main.go
+++ b/cmd/aws-janitor-boskos/main.go
@@ -25,8 +25,15 @@ import (
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
+
+	"k8s.io/test-infra/pkg/flagutil"
+	"k8s.io/test-infra/prow/config"
+	prowflagutil "k8s.io/test-infra/prow/flagutil"
 	"k8s.io/test-infra/prow/logrusutil"
+	prowmetrics "k8s.io/test-infra/prow/metrics"
+
 	"sigs.k8s.io/boskos/aws-janitor/account"
 	"sigs.k8s.io/boskos/aws-janitor/regions"
 	"sigs.k8s.io/boskos/aws-janitor/resources"
@@ -52,6 +59,19 @@ var (
 	includeTags common.CommaSeparatedStrings
 	excludeTM   resources.TagMatcher
 	includeTM   resources.TagMatcher
+
+	instrumentationOptions prowflagutil.InstrumentationOptions
+
+	cleaningTimeHistogram = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Name:        "aws_janitor_boskos_cleaning_time_seconds",
+		ConstLabels: prometheus.Labels{},
+		Buckets:     prometheus.ExponentialBuckets(1, 1.4, 30),
+	}, []string{"resource_type", "status"})
+
+	sweepsGauge = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name:        "aws_janitor_boskos_sweeps",
+		ConstLabels: prometheus.Labels{},
+	}, []string{"resource_type"})
 )
 
 const (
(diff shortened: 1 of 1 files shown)


cmd/aws-janitor-boskos/main.go
```diff
@@ -132,8 +164,12 @@ func cleanResource(res *common.Resource) error {
 		}
 	}
 
-	duration := time.Since(start)
-
-	logrus.WithFields(logrus.Fields{"name": res.Name, "duration": duration.Seconds()}).Info("Finished cleaning")
+	collectMetric(start, res.Name, "clean")
+	logrus.WithFields(logrus.Fields{"name": res.Name, "duration": time.Since(start).Seconds()}).Info("Finished cleaning")
```
ixdy, line 168: do you think we should include sweepCount somehow as well? might be interesting to understand if a longer duration is due to multiple sweeps.

cmd/aws-janitor-boskos/main.go
```diff
@@ -44,18 +51,33 @@ var (
...
+
+	instrumentationOptions prowflagutil.InstrumentationOptions
 )
 
 const (
 	sleepTime = time.Minute
 )
 
+var (
+	cleaningTimeHistogram = prometheus.NewHistogramVec(prometheus.HistogramOpts{
```
ixdy, line 63: nit: this could probably be included in the `var` block above.

ixdy (commented):
Thanks! This generally LGTM, just one question.

cmd/aws-janitor-boskos/main.go
cpanato: That sounds a great idea, I will update this PR over the weekend
thanks for your review and feedback

cmd/aws-janitor-boskos/main.go
cpanato: missed the var block above 🤦

cmd/aws-janitor-boskos/main.go
cpanato: added, but not 100% if it is correct, when you have time can you please take a look and let me know? thanks!

cmd/aws-janitor-boskos/main.go
```diff
@@ -116,24 +153,32 @@ func cleanResource(res *common.Resource) error {
...
 			time.Sleep(sweepSleepDuration)
 		}
 	}
 
-	duration := time.Since(start)
-
-	logrus.WithFields(logrus.Fields{"name": res.Name, "duration": duration.Seconds()}).Info("Finished cleaning")
+	sweepsGauge.WithLabelValues(res.Name).Set(float64(countSweeps))
+	collectMetric(start, res.Name, "clean")
+	logrus.WithFields(logrus.Fields{"name": res.Name, "duration": time.Since(start).Seconds(), "sweeps": countSweeps}).Info("Finished cleaning")
```
ixdy, line 177: Looking at this code again, I now realize I was misunderstanding it before - I thought that the specified `sweepCount` was the *maximum* number of attempted sweeps (i.e. retries if there was an error), but actually, I see that it always sweeps that many times, regardless of error or success.

We could keep the metric here, though it's not nearly as meaningful as I thought it was (since it'll basically remain the same for a given configuration). if we keep it, we can just pass `sweepCount` instead of using the separate `countSweeps` variable though.

cpanato: I can drop that for now and in a follow-up do a small refactor in this part to only retry when it failed, and then we add back the metric. WDTY?

ixdy: I think part of the reason it does multiple passes is that the AWS janitor typically does **not** return errors if it fails to delete resources, under the assumption that a later pass might successfully delete the resource (e.g. if there is some lingering dependency that gets cleaned up at a later stage).

This probably isn't ideal, but we'd need to investigate further and fix things to achieve the desired "retry only on error" behavior.

TL;DR we probably want to leave the existing behavior of always sweeping multiple times. The metrics you're adding can help us determine if this is taking too long and we should look into optimizing this process. :)

cpanato: @ixdy sounds good :)

update the code to reflect your feedback, thanks again

ixdy (approved):

PR #78, opened 2021-03-04
aws-janitor: add job duration metric

cpanato:
...to collect the response time when running this job

this is a short live job, and then we cannot guarantee that we will have someone to scrape the data at the correct time, to avoid that, we need to use the Prometheus PushGateway to store the metrics and then have Prometheus scrape the PushGateway, however, the problem continues, we need to make sure the metrics are pushed to the PushGateway in time.

So for that, we need to push ourselves, then the approach in this PR is to push when exiting the job

this is part of https://github.com/kubernetes-sigs/boskos/issues/13

opening this PR for feedback and see if sounds good

a screenshot for the metric pushed to PushGateway

Screen Shot 2021-03-04 at 14 00 38

/assign @ixdy

### DIFF
--- a/cmd/aws-janitor/main.go
+++ b/cmd/aws-janitor/main.go
@@ -52,11 +52,18 @@ var (
 	excludeTags common.CommaSeparatedStrings
 	includeTags common.CommaSeparatedStrings
 
+	sweepCount int
+
 	cleaningTimeHistogram = prometheus.NewHistogramVec(prometheus.HistogramOpts{
 		Name:        "aws_janitor_job_duration_time_seconds",
 		ConstLabels: prometheus.Labels{},
 		Buckets:     prometheus.ExponentialBuckets(1, 1.4, 30),
-	}, []string{"type", "status"})
+	}, []string{"type", "status", "region"})
+
+	sweepCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name:        "aws_janitor_swept_resources",
+		ConstLabels: prometheus.Labels{},
+	}, []string{"type", "status", "region"})
 )
 
 func init() {
@@ -81,13 +88,12 @@ func main() {
 	exitCode := 2
 	startProcess := time.Now()
 	if *pushGateway != "" {
-		jobName := fmt.Sprintf("aws-janitor-%s", common.GenerateDynamicResourceName())
 		registry := prometheus.NewRegistry()
-		registry.MustRegister(cleaningTimeHistogram)
-		pusher := push.New(*pushGateway, jobName).Gatherer(registry)
+		registry.MustRegister(cleaningTimeHistogram, sweepCounter)
+		pusher := push.New(*pushGateway, "aws-janitor").Gatherer(registry)
 
 		defer func() {
-			pushMetricBeforeExit(pusher, startProcess, *cleanAll, exitCode)
+			pushMetricBeforeExit(pusher, startProcess, exitCode)
 			os.Exit(exitCode)
 		}()
 	} else {
@@ -174,17 +180,17 @@ func markAndSweep(opts resources.Options, region string) error {
 		}
 	}
 
-	swept := res.MarkComplete()
+	sweepCount = res.MarkComplete()
 	if err := res.Save(opts.Session, s3p); err != nil {
 		return errors.Wrapf(err, "Error saving %q", *path)
 	}
 
-	logrus.Infof("swept %d resources", swept)
+	logrus.Infof("swept %d resources", sweepCount)
 
 	return nil
 }
 
-func pushMetricBeforeExit(pusher *push.Pusher, startTime time.Time, jobType bool, exitCode int) {
+func pushMetricBeforeExit(pusher *push.Pusher, startTime time.Time, exitCode int) {
 	// Set the status of the job
 	status := "failed"
 	if exitCode == 0 {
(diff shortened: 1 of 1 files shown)


cmd/aws-janitor/main.go
```diff
@@ -143,3 +183,24 @@ func markAndSweep(opts resources.Options, region string) error {
...
 }
+
+func pushMetricBeforeExit(pusher *push.Pusher, startTime time.Time, jobType bool, exitCode int) {
+	// Set the status of the job
+	status := "failed"
+	if exitCode == 0 {
+		status = "success"
+	}
+
+	// Set the type of the job to report the metric
```
ixdy, line 194: I think a better label here might be `clean_all` which is then either set to `true` or `false` depending on the flag setting.

cmd/aws-janitor/main.go
```diff
@@ -34,16 +40,23 @@ import (
...
+	pushGateway = flag.String("push-gateway", "", "If specified, push prometheus metrics to this endpoint.")
 
 	excludeTags common.CommaSeparatedStrings
 	includeTags common.CommaSeparatedStrings
+
+	cleaningTimeHistogram = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Name:        "aws_janitor_job_duration_time_seconds",
+		ConstLabels: prometheus.Labels{},
+		Buckets:     prometheus.ExponentialBuckets(1, 1.4, 30),
+	}, []string{"type", "status"})
```
ixdy, line 59: see note below about `type`, but I think it might also be worth including `region`

cmd/aws-janitor/main.go
ixdy: (also, since it's a global variable, you may not need to pass along the setting of `cleanAll`)

cmd/aws-janitor/main.go
```diff
@@ -143,3 +183,24 @@ func markAndSweep(opts resources.Options, region string) error {
```
ixdy, line 183: no action required in this change, but in the future it might be worth reporting how many resources were deleted, too (see line directly above this one, I just can't comment on it directly)

(there's probably a similar count obtainable from `CleanAll()`)

ixdy (commented):
thanks for this PR! just a few minor suggestions, otherwise this largely LGTM

cmd/aws-janitor/main.go
```diff
@@ -143,3 +183,24 @@ func markAndSweep(opts resources.Options, region string) error {
...
+	}
+
+	// Set the type of the job to report the metric
+	job := "sweep"
+	if jobType {
+		job = "clean"
+	}
+
+	duration := time.Since(startTime).Seconds()
+	cleaningTimeHistogram.WithLabelValues(job, status).Observe(duration)
```
ixdy, line 201: also, slight preference for using `With(prometheus.Labels{"key": value, ...})` over `WithLabelValues`, as the latter feels a bit more brittle.

cmd/aws-janitor/main.go
```diff
@@ -63,24 +76,47 @@ func main() {
 	}
 	logrus.SetLevel(level)
 
+	// If Prometheus PushGateway is configured, then before exit push the metric
+	// to the PushGateway instance, otherwise just exit
+	exitCode := 2
+	startProcess := time.Now()
+	if *pushGateway != "" {
+		jobName := fmt.Sprintf("aws-janitor-%s", common.GenerateDynamicResourceName())
```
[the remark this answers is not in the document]
cpanato: i was thinking and maybe not, we can have the same name

cmd/aws-janitor/main.go
cpanato: makes sense, thanks for the feedback

cmd/aws-janitor/main.go
cpanato: ok!

cmd/aws-janitor/main.go
cpanato: thanks for your review and feedback Jeff

Added a metric to report the sweep, the `CleanAll()`  function does not return how many resources were deleted, but i think we can add that, i can do that if you think that will be good

ptal

cmd/aws-janitor/main.go
```diff
@@ -34,16 +40,30 @@ import (
...
+
+	sweepCount int
+
+	cleaningTimeHistogram = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Name:        "aws_janitor_job_duration_time_seconds",
+		ConstLabels: prometheus.Labels{},
+		Buckets:     prometheus.ExponentialBuckets(1, 1.4, 30),
+	}, []string{"type", "status", "region"})
+
+	sweepCountGauge = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
ixdy, line 63: it seems like maybe this should be a counter instead of a gauge - what do you think?

cmd/aws-janitor/main.go
```diff
@@ -34,16 +40,30 @@ import (
...
+		Name:        "aws_janitor_sweeps",
```
ixdy, line 64: nit: maybe `aws_janitor_swept_resources`? `sweeps` makes me think of how many times the sweep operation was run, rather than the number of items swept.

cmd/aws-janitor/main.go
ixdy: yeah, we'd need to pass that information back from `CleanAll()`, but that can happen separately.

cmd/aws-janitor/main.go
```diff
@@ -134,12 +180,41 @@ func markAndSweep(opts resources.Options, region string) error {
...
+	// Set the status of the job
+	status := "failed"
+	if exitCode == 0 {
+		status = "success"
+	}
+
+	// Set the type of the job to report the metric
+	var job string
+	if !*cleanAll {
+		job = "sweep"
```
ixdy, line 203: nit: maybe `mark_and_sweep`?

cmd/aws-janitor/main.go
cpanato: again, thanks for your time to review this PR

ixdy (approved):

PR #87, opened 2021-06-19
Oneshot mode for aws-janitor

chemikadze:
In some cases it is useful to run one-off cleanup of
existing resources without relying on s3 state, but
still honoring TTLs.

Also clarifies that -all ignores TTLs.


chemikadze:
cc @ixdy

ixdy (commented):
Have you test this in dry-run mode? Does it seem to do the right thing? (lacking any better e2e tests)

Not sure if it's worth noting that this won't really work with some resources (e.g. VPCs), which have no creation time metadata.

I'm also now wondering whether the `-all` and `-oneshot` flows even need to call the `ListAll` functions first - it seems like setting the TTL on an empty set and then passing it to the `MarkAndSweep` functions should adequately delete things now, since `set.Mark` knows how to find a relevant creation timestamp for most resources.

PR #90, opened 2021-07-07
Replace zones flag with additional filter

azylinski:
Fix WARNING:

Flag `--zones` is deprecated. Use `--filter="zone:( ZONE ... )"` instead.
For example `--filter="zone:( europe-west1-b europe-west1-c )"`.

### DIFF
--- a/cmd/janitor/gcp_janitor.py
+++ b/cmd/janitor/gcp_janitor.py
@@ -256,14 +256,17 @@ def collect(project, age, resource, filt, clear_all):
     if resource.name == 'sinks' and not clear_all:
         return col
 
+    if resource.condition == 'zone' and resource.name not in ['sole-tenancy', 'network-endpoint-groups']:
+        if len(filt) > 0:
+            filt.append(' AND ')
+        filt.append('zone=(%s)' % ','.join(ZONES))
+
     cmd = base_command(resource)
     cmd.extend([
         'list',
         '--format=json(name,creationTimestamp.date(tz=UTC),createTime.date(tz=UTC),zone,region,isManaged)',
-        '--filter=%s' % filt,
+        '--filter="%s"' % filt,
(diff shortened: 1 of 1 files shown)


alvaroaleman (commented):
Deferring to Jeff, I am not really familiar with the janitors
/ok-to-test
/uncc

cmd/janitor/gcp_janitor.py
```diff
@@ -256,14 +256,17 @@ def collect(project, age, resource, filt, clear_all):
     if resource.name == 'sinks' and not clear_all:
         return col
 
+    if resource.condition == 'zone' and resource.name not in ['sole-tenancy', 'network-endpoint-groups']:
+        if len(filt) > 0:
+            filt.append(' AND ')
+        filt.append('zone=(%s)' % ','.join(ZONES))
```
ixdy, line 262: are these supposed to be comma-separated or space-separated?

cmd/janitor/gcp_janitor.py
```diff
@@ -256,14 +256,17 @@ def collect(project, age, resource, filt, clear_all):
     if resource.name == 'sinks' and not clear_all:
         return col
 
+    if resource.condition == 'zone' and resource.name not in ['sole-tenancy', 'network-endpoint-groups']:
+        if len(filt) > 0:
+            filt.append(' AND ')
```
ixdy, line 261: isn't `filt` a string? don't you want `+=` here (and the next line)?

ixdy (commented):
Have you tested this change locally? Our automated testing doesn't really test the janitor.

Does this work with all resource types? I'm concerned whether there are any that only support `--zones` and not specifying the `zone` in the filter.

/ok-to-test

ixdy:
/close

I think this was obsoleted by #98.

Issue #92, opened 2021-07-19
Handwritten CRD DeepCopy methods are not deepcopying


alvaroaleman:
For some reason we have handwritten instead of generated DeepCopy methods on our CRDs and they definitely do not DeepCopy, i.E. something like this is incorrect, because nested pointers will just get copied over, so if someone changes the value that is pointed to, it will change on both the original and the "deepcopied" version:

https://github.com/kubernetes-sigs/boskos/blob/36bb085667e74f1e79c7d18ab4446216c9e2f2be/crds/drlc_crd.go#L76

/kind bug

PR #93, opened 2021-07-21
Generate DeepCopy methods for CRD types

alvaroaleman:
The handwritten ones were buggy, it is much safer to generate these.

Fixes https://github.com/kubernetes-sigs/boskos/issues/92


Makefile
```diff
@@ -88,6 +92,16 @@ verify-boilerplate:
 verify-lint: $(GOLANGCI_LINT)
 	./hack/tools/bin/golangci-lint run -v --new-from-rev HEAD~
 
+.PHONY: verify-codegen
+verify-codegen: $(CONTROLLER_GEN)
```
ixdy, line 96: maybe add this to the verify rule?

ixdy (commented):
The CI failure looks legitimate, too.

Makefile
alvaroaleman: Done

alvaroaleman:
@ixdy updated the PR, it is now bigger because the DeepCopy generation requires all sub-types to also have a DeepCopy which means we can not use any stdlib types in our api, so I had to replace `time.Time` with `metav1.Time` and the `Userdata` map which is a wrapper around `sync.Map` with a normal `map[string]string` which in turn required a bunch of additional changes all over the place.

common/common.go
```diff
@@ -261,9 +261,9 @@ func (ud *UserData) Set(id string, in interface{}) error {
 
 // Update updates existing UserData with new UserData.
 // If a key as an empty string, the key will be deleted
-func (ud *UserData) Update(new *UserData) {
+func (ud *UserData) Update(new *UserData) *UserData {
```
stevekuznetsov, line 264: Could we add godoc as to why this returns a pointer? Isn't it self-mutating as well?

alvaroaleman: It returns a pointer so it can be chained, i.E. `ud.Update(otherUD).ToMap()`

handlers/server_client_test.go
```diff
@@ -190,6 +190,7 @@ func TestAcquireByState(t *testing.T) {
 		},
 	}
 	for _, tc := range testcases {
+		tc := tc
```
ixdy, line 193: why does lint need this?

crds/resource_crd.go
```diff
@@ -121,39 +105,22 @@ func FromResource(r common.Resource) *ResourceObject {
 		Status: ResourceStatus{
 			Owner:          r.Owner,
 			State:          r.State,
-			LastUpdate:     r.LastUpdate,
-			UserData:       r.UserData,
-			ExpirationDate: r.ExpirationDate,
+			LastUpdate:     v1.Time{Time: r.LastUpdate.UTC()},
```
ixdy, line 108: why does `LastUpdate` use UTC but `ExpirationDate` (through `timeToMetaTime()`) does not?

ranch/storage.go
```diff
@@ -108,7 +108,7 @@ func (s *Storage) GetResource(name string) (*crds.ResourceObject, error) {
 		return nil, fmt.Errorf("failed to get resource %s: %v", name, err)
 	}
 	if o.Status.UserData == nil {
-		o.Status.UserData = &common.UserData{}
+		o.Status.UserData = map[string]string{}
```
ixdy, line 111: nitful suggestion: I think we have a `common.UserDataMap` type that you could use here (and elsewhere) instead of `map[string]string`.

ixdy (commented):
generally LGTM. curious if @stevekuznetsov has any other comments.

handlers/server_client_test.go
alvaroaleman: Because of https://github.com/kyoh86/scopelint/issues/4

Scopelint is deprecated though, we might want to replace it in a followup with exportloopref

crds/resource_crd.go
alvaroaleman: That was a remainder, removed it

ranch/storage.go
alvaroaleman: the `common.UserDataMap` is aliasing a `map[string]string` without adding any methods which IMHO makes the code that uses it harder to understand without adding any benefit that I can see. I'd rather remove `common.UserDataMap` in a follow-up.

stevekuznetsov:
This LGTM from me
/lgtm
/hold cancel

PR #95, opened 2021-07-30
Update test-infra and use global secret agent

alvaroaleman:
Test-Infras secret agent sets a logrus formatter with ultimatively is a package-level variable which causes tests that use it to race. To avoid this, we made it a global, for more context see https://github.com/kubernetes/test-infra/pull/23063

This change updates boskos test-infra dependency and makes it use the new global secret agent, so we can then in test-infra update its boskos dependency and remove the non-global agent.


ixdy (commented):
/lgtm2

PR #97, opened 2021-08-09
Add cleaner for libvirt CIs

hamzy:
This adds a cleaner for CIs using a libvirt deployment.  After a job has
run, cleanup of libvirt resources needs to be enforced before another
libvirt deployment can happen.

### DIFF
--- a/cmd/libvirt-cleaner/cleaner.go
+++ b/cmd/libvirt-cleaner/cleaner.go
@@ -1,0 +1,181 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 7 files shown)


ixdy:
I'm not particularly familiar with libvirt - does it make sense to implement this as a cleaner, or should this be implemented as a janitor?

alvaroaleman:
> I'm not particularly familiar with libvirt - does it make sense to implement this as a cleaner, or should this be implemented as a janitor?

I think a janitor makes more sense. Also I am not sure if this is generic enough to make sense upstream? This seems super RedHat-specific. If we want it it here, all of that would need to become config options.

Resources have btw types, which is how you infer the type, rather than using its name.

PR #98, opened 2021-08-10
[GCP janitor] Update ZONES and use the new filter syntax

chizhg:
1. Update `ZONES` list for GCP janitor by running `gcloud compute zones list --format="value(name)" | sort | awk '{print "    \x27"$1"\x27," }'`
2. Fix "WARNING: Flag `--zones` is deprecated. Use `--filter="zone:( ZONE ... )"` instead."

/cc @ixdy

### DIFF
--- a/cmd/janitor/gcp_janitor.py
+++ b/cmd/janitor/gcp_janitor.py
@@ -110,6 +110,9 @@ ZONES = [
     'asia-south1-a',
     'asia-south1-b',
     'asia-south1-c',
+    'asia-south2-a',
+    'asia-south2-b',
+    'asia-south2-c',
     'asia-southeast1-a',
     'asia-southeast1-b',
     'asia-southeast1-c',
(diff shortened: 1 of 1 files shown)


cmd/janitor/gcp_janitor.py
```diff
@@ -147,6 +147,7 @@
     'us-central1-b',
     'us-central1-c',
     'us-central1-f',
+    'us-east1-a',
```
ixdy, line 150: `us-east1-a` probably didn't exist when this list was populated.

Are there any other zones that should be added? See the comment on L93 describing how this list was generated.

ixdy (approved):
/hold

Have you tested this in dry-run mode?

cmd/janitor/gcp_janitor.py
chizhg: Hmm I tried running the command on L93, it did generate a few more zones but did not include `us-east1-a`. 

After I added `us-east1-a` in this list, I did try running the script against a project that has GCE instances in the `us-east1-a` zone, and it worked.

chizhg: Per offline discussion, us-east1-a is a private zone and is not exposed to all projects, so removing it and update the `ZONES` list with the gcloud command instead.

ixdy (approved):
/hold cancel

PR #99, opened 2021-08-16
Boskos: Allow HA by adding optional leader election

alvaroaleman:
This change optionally makes the Boskos server use leader election. This
allows to have multiple replicas running of which only one is active in
order to reduce downtime when one replica gets unavailable.

### DIFF
--- a/cmd/boskos/boskos.go
+++ b/cmd/boskos/boskos.go
@@ -29,10 +29,14 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/viper"
 	corev1 "k8s.io/api/core/v1"
+	toolsleaderelection "k8s.io/client-go/tools/leaderelection"
+	"k8s.io/client-go/tools/leaderelection/resourcelock"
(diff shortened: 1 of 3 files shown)


cmd/boskos/boskos.go
```diff
@@ -117,6 +123,46 @@ func main() {
 		logrus.WithError(err).Fatal("Failed to get mgr")
 	}
 
+	if *enableLeaderElection {
```
stevekuznetsov, line 126: is this guaranteed to not be `nil`?

stevekuznetsov (approved):
/hold

LGTM - @ixdy do you want to look?

cmd/boskos/boskos.go
alvaroaleman: Yeah, the flag gurantees that

stevekuznetsov:
Ping @ixdy - I'll unhold tomorrow if you don't have any outstanding comments, this will be nice to have for redeploy

PR #101, opened 2021-08-20
Allow specifying additional zones for GCP janitor

chizhg:
Fixes https://github.com/kubernetes-sigs/boskos/issues/100

/cc @ixdy

PR #102, opened 2021-08-20
Update verify_boilerplate to python3

ixdy:
As seen in #101, `verify_boilerplate.py` has started failing.

I believe this is because the `golang:1.16` image is now based on Debian Bullseye, which I guess has deprecated the `python` alias (finally).

I ran `2to3 -p -e -w hack/verify/verify_boilerplate.py` to update this script.

/assign @chizhg


ixdy:
/hold

This might not be working...

ixdy:
/hold cancel

OK I think this actually works again.

PR #103, opened 2021-09-09
Boskos: Manage a label that denotes leader to allow HA setups

alvaroaleman:
This change adds a small controller to boskos that will make it manage a
label on pods. The label will be set on the pod that has the leader
lease and removed on all other pods. This in turn allows to have
multiple replicas of boskos running by using a service that targets said
label.

It makes the new `--boskos-label-selector` and `--pod-name` flags mandatory when `--enable-leader-election` is set.

Apart from the included unit tests I also manually deployed and tested this with kind:

```
$ k get pod -o wide
NAME                      READY   STATUS    RESTARTS   AGE   IP            NODE                 NOMINATED NODE   READINESS GATES
boskos-5dfb47b97b-gsgcx   1/1     Running   0          18m   10.244.0.26   kind-control-plane   <none>           <none>
boskos-5dfb47b97b-hbcd8   1/1     Running   0          17m   10.244.0.27   kind-control-plane   <none>           <none>
boskos-5dfb47b97b-ppbjv   1/1     Running   0          18m   10.244.0.25   kind-control-plane   <none>           <none>

$ k get endpoints
NAME             ENDPOINTS                                            AGE
boskos           10.244.0.27:8080                                     61m
boskos-metrics   10.244.0.25:9090,10.244.0.26:9090,10.244.0.27:9090   61m
```

### DIFF
--- a/cmd/boskos/boskos.go
+++ b/cmd/boskos/boskos.go
@@ -29,14 +29,11 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/viper"
 	corev1 "k8s.io/api/core/v1"
-	toolsleaderelection "k8s.io/client-go/tools/leaderelection"
-	"k8s.io/client-go/tools/leaderelection/resourcelock"
-	"k8s.io/client-go/tools/record"
+	"k8s.io/apimachinery/pkg/labels"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
-	"sigs.k8s.io/controller-runtime/pkg/leaderelection"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
@@ -54,6 +51,7 @@ import (
 	"sigs.k8s.io/boskos/common"
 	"sigs.k8s.io/boskos/crds"
 	"sigs.k8s.io/boskos/handlers"
+	"sigs.k8s.io/boskos/leaderlabelreconciler"
 	"sigs.k8s.io/boskos/metrics"
 	"sigs.k8s.io/boskos/ranch"
 )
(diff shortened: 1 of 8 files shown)


stevekuznetsov:
/cc @ixdy @cjwagner

This is an approach we've seen implemented to reasonable success in other projects (like Vault) which allows for a singleton to run HA and have a) zero-downtime upgrades and b) canary deployments

cmd/boskos/boskos.go
```diff
@@ -68,12 +66,14 @@ var (
 	configPath = flag.String("config", "config.yaml", "Path to init resource file")
 	_          = flag.Duration("dynamic-resource-update-period", defaultDynamicResourceUpdatePeriod,
 		"Legacy flag that does nothing but is kept for compatibility reasons")
-	requestTTL              = flag.Duration("request-ttl", defaultRequestTTL, "request TTL before losing priority in the queue")
-	logLevel                = flag.String("log-level", "info", fmt.Sprintf("Log level is one of %v.", logrus.AllLevels))
-	namespace               = flag.String("namespace", corev1.NamespaceDefault, "namespace to install on")
-	port                    = flag.Int("port", 8080, "Port to serve on")
-	enableLeaderElection    = flag.Bool("enable-leader-election", false, "If leader election should be enabled. If enabled, it is possible to run Boskos with multiple replicas and only the one that is leader will receive traffic.")
```
cjwagner, line 75: So did this not work previously?

cmd/boskos/boskos.go
```diff
@@ -68,12 +66,14 @@ var (
...
+	requestTTL           = flag.Duration("request-ttl", defaultRequestTTL, "request TTL before losing priority in the queue")
+	logLevel             = flag.String("log-level", "info", fmt.Sprintf("Log level is one of %v.", logrus.AllLevels))
+	namespace            = flag.String("namespace", corev1.NamespaceDefault, "namespace to install on")
+	port                 = flag.Int("port", 8080, "Port to serve on")
+	enableLeaderElection = flag.Bool("enable-leader-election", false, "If leader election should be enabled. If enabled, it is possible to run Boskos with multiple replicas and only the one that is leader will receive traffic. This works through boskos managing a label so it is only set on the leader pod. The boskos service must select on this label and not one that matches all replicas.")
+	boskosLabelSelector  = flag.String("boskos-label-selector", "", "Must be set when leader election is used. The label selector to identify boskos pods")
+	podName              = flag.String("pod-name", "", "The name of the current pod. Required when --enable-leadere-election is set.")
```
cjwagner, line 75: This is a breaking change for existing boskos instances using leader election. We should probably avoid this by retaining the old behavior unless opted in by specifying these flags.

cmd/boskos/boskos.go
```diff
@@ -118,49 +118,29 @@ func main() {
...
-			LeaseDuration: 15 * time.Second,
-			RenewDeadline: 10 * time.Second,
-			RetryPeriod:   2 * time.Second,
-			Callbacks: toolsleaderelection.LeaderCallbacks{
-				OnStartedLeading: func(_ context.Context) { close(elected) },
-				OnStoppedLeading: func() { logrus.Fatal("Lost leaderelection") },
-			},
-			ReleaseOnCancel: true,
-		})
+		fmt.Printf("label selector: %s\n", *boskosLabelSelector)
```
cjwagner, line 133: Left over from debugging? Otherwise use `logrus`.

cjwagner (commented):
> @stevekuznetsov: GitHub didn't allow me to request PR reviews from the following users: cjwagner.

I'm not sure if I can LGTM in this repo.

I'm not familiar with the boskos source, but the overall change seems fine to me. I didn't do a close review, but did notice that this looks like a backwards incompatible change and I think we should try to avoid that.

cmd/boskos/boskos.go
alvaroaleman: It did work but it broke upgrades when using it, as only one replica would get ready (the one that had the leader lease) and the others would not which in turn makes the deployment controller halt rollout.

cmd/boskos/boskos.go
alvaroaleman: The leader election mechanism was only added very recently in https://github.com/kubernetes-sigs/boskos/pull/99, didn't solve the problem it was supposed to solve and this only breaks ppl if they opted in  - I would argue that this is unlikely to break anyone.

spiffxp (commented):
I'm not aware of any deployments for kubernetes that used the old `--enable-leader-election` flag (ref: https://cs.k8s.io/?q=%2Fboskos%3A&i=nope&files=.*yaml&excludeFiles=&repos=)

@cjwagner are you aware of any others that might use this?

If you update the PR description with what someone using the old flags would need to change their invocation to instead, I'm OK seeing this merge

alvaroaleman:
> If you update the PR description with what someone using the old flags would need to change their invocation to instead, I'm OK seeing this merge

Updated the description

cjwagner:
> @cjwagner are you aware of any others that might use this?

No I am not. I think it is unlikely that there are any given that this is a relatively new flag: https://github.com/kubernetes-sigs/boskos/pull/103#discussion_r706402899

spiffxp (approved):
/approve
/lgtm

Issue #94, opened 2021-07-28
Boskos CRD uses deprecated API group to be removed in v1.22


spiffxp:
https://kubernetes.io/docs/reference/using-api/deprecation-guide/#customresourcedefinition-v122

>The **apiextensions.k8s.io/v1beta1** API version of CustomResourceDefinition will no longer be served in v1.22.
>
>* Migrate manifests and API clients to use the **apiextensions.k8s.io/v1** API version, available since v1.16.
>* All existing persisted objects are accessible via the new API
>* Notable changes:
>    * `spec.scope` is no longer defaulted to `Namespaced` and must be explicitly specified
>    * `spec.version` is removed in v1; use `spec.versions` instead
>    * `spec.validation` is removed in v1; use `spec.versions[*].schema` instead
>    * `spec.subresources` is removed in v1; use `spec.versions[*].subresources` instead
>    * `spec.additionalPrinterColumns` is removed in v1; use `spec.versions[*].additionalPrinterColumns` instead
>    * `spec.conversion.webhookClientConfig` is moved to `spec.conversion.webhook.clientConfig` in v1
>    * `spec.conversion.conversionReviewVersions` is moved to `spec.conversion.webhook.conversionReviewVersions` in v1
>    * `spec.versions[*].schema.openAPIV3Schema` is now required when creating v1 CustomResourceDefinition objects, and must be a [structural schema](/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#specifying-a-structural-schema)
>    * `spec.preserveUnknownFields: true` is disallowed when creating v1 CustomResourceDefinition objects; it must be specified within schema definitions as `x-kubernetes-preserve-unknown-fields: true`
>    * In `additionalPrinterColumns` items, the `JSONPath` field was renamed to `jsonPath` in v1 (fixes [#66531](https://github.com/kubernetes/kubernetes/issues/66531))

spiffxp:
FYI @alvaroaleman @stevekuznetsov

I see you're working on CRD-related stuff right now, wasn't sure if this had popped up on your radar yet

spiffxp:
I anticipate the k8s-infra prow clusters are going to be up to v1.22 before the end of the year, and we'll want this handled by then

spiffxp:
I've updated k8s-infra-prow-build via https://github.com/kubernetes/k8s.io/pull/3050

Will let this soak for a bit before deciding to update other clusters

PR #105, opened 2021-09-28
Update CRDs to apiextensions.k8s.io/v1

spiffxp:
Related:
- https://github.com/kubernetes-sigs/boskos/issues/94

These are the changes necessary to bring the CRD definition in line with changes called out in https://kubernetes.io/docs/reference/using-api/deprecation-guide/#customresourcedefinition-v122

WIP because:
- needs upgrade testing

The biggest change here is that the schema is now required. I wrote this
by looking at the types in crd/ and common/. Any field that was a map
got a `x-kubernetes-preserve-unknown-fields: true`

The most controversial change is probably adding the api-approved
annotation, to allow us to keep using the .k8s.io group. This should
mean that updating the CRD doesn't require migrating any of the
resources. As prior art, I'm looking at the fact that prow.k8s.io
did the same thing for its ProwJob CRD.

This is controversial because these CRD's aren't really part of core
kubernetes. They should be moved to the x-k8s.io group. However it's not
immediately clear to me what it would take to have boskos support
transparently migrating resources from CRD's define in one group to
another.

This has been tested by running the example deployment against a kind
cluster running kubernetes 1.22

### DIFF
--- a/deployments/base/cleaner/deployment.yaml
+++ b/deployments/base/cleaner/deployment.yaml
@@ -33,7 +33,7 @@ spec:
       terminationGracePeriodSeconds: 300
       containers:
         - name: boskos-cleaner
-          image: gcr.io/k8s-staging-boskos/cleaner:v20200617-f289ba6
+          image: gcr.io/k8s-staging-boskos/cleaner:v20211015-2401f5c
           args:
             - --boskos-url=http://boskos
             - --namespace=$(NAMESPACE)
--- a/deployments/base/crd.yaml
+++ b/deployments/base/crd.yaml
@@ -12,10 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: dynamicresourcelifecycles.boskos.k8s.io
+  annotations:
+    api-approved.kubernetes.io: https://github.com/kubernetes-sigs/boskos/pull/105
 spec:
   group: boskos.k8s.io
   names:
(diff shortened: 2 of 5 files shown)


spiffxp:
/priority important-soon
We've got boskos running on clusters that just got alerted a 1.21 upgrade is pending, so it's time to start thinking about the upgrade story before we're forced to

ixdy:
/lgtm
/hold
in case anyone else wants to comment.

What does the upgrade story look like here? Can you just seamlessly apply this, or do you have to delete the old CRD spec and apply the new one? (What happens to all of the existing resources in the latter case?)

ixdy:
also, have you tested these new specs in a deployment yet?

alvaroaleman:
This doesn't appear to quite work yet:

```
$ k apply -f deployments/base/crd.yaml
Error from server (Invalid): error when creating "deployments/base/crd.yaml": CustomResourceDefinition.apiextensions.k8s.io "dynamicresourcelifecycles.boskos.k8s.io" is invalid: [spec.versions[0].schema.openAPIV3Schema: Required value: schemas are required, metadata.annotations[api-approved.kubernetes.io]: Required value: protected groups must have approval annotation "api-approved.kubernetes.io", see https://github.com/kubernetes/enhancements/pull/1111]
Error from server (Invalid): error when creating "deployments/base/crd.yaml": CustomResourceDefinition.apiextensions.k8s.io "resources.boskos.k8s.io" is invalid: [spec.versions[0].schema.openAPIV3Schema: Required value: schemas are required, metadata.annotations[api-approved.kubernetes.io]: Required value: protected groups must have approval annotation "api-approved.kubernetes.io", see https://github.com/kubernetes/enhancements/pull/1111]
```

spiffxp:
OK I took a more honest crack at this today.  Description updated with the changes made.

I did a limited amount of manual upgrade testing with the example deployment:
- stood up a kind cluster at v1.21.1 with the old CRD
- used `kubectl port-forward` and ran `boskosctl acquire` a few times from my laptop to check out resources
- `kubectl apply`'ed the new CRD
- verified existing `resources` and `dynamicresourcelifecycles` were still listable via `kubectl get`
- verified `boskosctl` was able to release resources it had previously acquired
- verified `boskosctl` was able to acquire resources

I personally am comfortable enough based on this to try applying the new CRD to the k8s-infra-prow-build cluster next week.  But I also understand the level of comfort "it worked on my machine" might provide to folks who have a more complicated setup.

Two questions:
- Are we ok proceeding with keeping these CRDs in a group ending in k8s.io as was done for prow?
- What other testing do you think we should do before I drop the WIP tag from this?

spiffxp:
I've deployed to k8s-infra-prow-build via https://github.com/kubernetes/k8s.io/pull/3050

spiffxp:
/cc @chaodaiG @cjwagner 
FYI, in case you have any input for the above questions, i.e. what more would you want to see before rolling this change out to k8s-prow-builds

alvaroaleman (approved):
looks good, holding in case chao or cole have comments
/hold

cjwagner (approved):
No objections from me, looks good.

PR #111, opened 2021-11-16
Add filestore instance support to GCP janitor

chemikadze:


cmd/janitor/gcp_janitor.py
```diff
@@ -283,7 +288,9 @@ def collect(project, zones, age, resource, filt, clear_all):
         'list',
         '--format=json(name,creationTimestamp.date(tz=UTC),createTime.date(tz=UTC),zone,region,isManaged)',
         '--project=%s' % project])
-    if resource.condition == 'zone' and resource.name != 'sole-tenancy' and resource.name != 'network-endpoint-groups':
+    if (resource.condition == 'zone'
```
chizhg, line 291: What's the reason for deleting `resource.name != 'network-endpoint-groups'`? Was it by mistake?

cmd/janitor/gcp_janitor.py
```diff
@@ -318,6 +325,10 @@ def collect(project, zones, age, resource, filt, clear_all):
                 # subcommands want just the name, not the full URL, so strip it.
                 colname = item[resource.condition].rsplit('/', 1)[-1]
                 log('looking for items in %s=%s' % (resource.condition, colname))
+            elif resource.group == 'filestore':
```
chizhg, line 328: I'm not familiar with `filestore` so not entirely sure how it works. Were you able to run this script against your own project to make sure this does not break?

cmd/janitor/gcp_janitor.py
chemikadze: Great catch, thanks! Wanted to move per-resource match to separate lines, lost during reformatting.

cmd/janitor/gcp_janitor.py
chemikadze: Yes, I have verified it with my project having filestore instances.

chemikadze:
@stevekuznetsov @alvaroaleman do you mind taking a look at the PR? Thanks in advance!

alvaroaleman (approved):

PR #131, opened 2022-07-04
Bump test infra dependency

olemarkus:


cjwagner (commented):
Could you please share some context/motivation about this update and all the new `require` statements? Go mod files still confuse me a bit.

olemarkus:
Sure. The short version is that go automatically adds all indirect 1.17 modules as an additional block in `go.mod`. This list is exhaustive vs before when it only contained indirect dependencies one wanted a specific version of. The longer version is here https://go.dev/doc/go1.17#go-command

I.e it is just a consequence of bumping the go version in the `go.mod file`.

Issue #132, opened 2022-07-07
aws-janitor overly eagerly deleting IAM Role


justinsb:
On AWS, IAM Roles have the same name (and no unique UUID).  They do have a creation timestamp, however.

Our test jobs are creating IAM roles with the same name.  The aws-janitor runs periodically, and if the timings work out "just so", aws-janitor will observe different IAM roles with the same name for the entire TTL window.  It will then delete an IAM role, thinking that it is no longer in use, but in fact it has seen multiple different IAM roles with the same name.

I propose using CreationTimestamp to differentiate.

olemarkus:
@justinsb is this what happens?
My observation is that the janitor runs between IAM role created and IAM role used. That's also why only the pod roles are deleted, as the instance roles are used quickly after cluster creation, while pod roles are used 5-10 minutes later.

justinsb:
There is a supplemental check that the IAM role isn't in use: https://github.com/kubernetes-sigs/boskos/blob/master/aws-janitor/resources/iam_roles.go#L89-L92

This may be what stops the instance roles from being deleted.

I do think the logic is wrong though, if we have a creationTimestamp from AWS we should treat that as the creation timestamp, not some other value that we have constructed ourselves :-)

olemarkus:
Yeah. The instance roles are used earlier, so they get protection from that supplemental check, while the Pod ServiceAccount roles do not because they are not used yet.

I agree that if we already have a creation time check we should just use this. However, we are still seeing this after your PR merged.

https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-aws-k8s-latest/1556000917520650240/build-log.txt

https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-aws-k8s-latest/1556000917520650240/artifacts/cluster-info/kube-system/dns-controller-55fd4d7694-t4ts2/logs.txt

olemarkus:
So we actually had a leak as well: https://github.com/kubernetes/kops/pull/14151

Let's wait for this one to merge and observe the e2es.

olemarkus:
I think we can close this one now.

PR #133, opened 2022-07-07
aws-janitor: Use the creation time if it is specified

justinsb:
This stops us deleting objects that are repeatedly recreated with the
same name, as long as the creation time is passed in.

Issue #132

### DIFF
--- a/aws-janitor/resources/set_test.go
+++ b/aws-janitor/resources/set_test.go
@@ -125,11 +125,14 @@ func TestMarkCreationTimes(t *testing.T) {
 				ExpectedFirstSeen: threeHoursAgo,
 			},
 			{
-				// Some resources only have times that can potentially reset to newer times (e.g. startup date)
+				// Some resources might be created again with the same name, and we don't want to delete those.
+				// Note that some resources don't record a creation timestamp,
+				// and only have times that can potentially reset to newer times (e.g. startup date),
+				// potentially we will leak those, but if the data keeps increasing that is taken as a usage signal.
 				Resource:          prevSeenNewYoungerCreateTime,
-				ShouldDelete:      true,
+				ShouldDelete:      false,
 				CreateTime:        &tenMinutesAgo,
-				ExpectedFirstSeen: threeHoursAgo,
+				ExpectedFirstSeen: tenMinutesAgo,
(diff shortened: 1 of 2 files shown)


aws-janitor/resources/set_test.go
```diff
@@ -125,11 +125,14 @@ func TestMarkCreationTimes(t *testing.T) {
 				ExpectedFirstSeen: threeHoursAgo,
 			},
 			{
-				// Some resources only have times that can potentially reset to newer times (e.g. startup date)
+				// Some resources might be created again with the same name, and we don't want to delete those.
+				// Note that some resources don't record a creation timestamp,
+				// and only have times that can potentially reset to newer times (e.g. startup date),
+				// potentially we will leak those, but if the data keeps increasing that is taken as a usage signal.
```
cjwagner, line 131: I don't follow this last part. Why are we ok with leaking these boskos resources?

justinsb: So I'd argue that we shouldn't pass creationTimestamp unless it is a creation timestamp, but that felt like a bigger change.

Right now, there are a few places where pass a different value instead of nil or the creationTimestamp:

* Instances pass the LaunchTime
* Disks pass the AttachTime
* Disk Snapshots pass the StartTime
* SQS passes what looks like a creation timestamp, but parsed from an attribute


These values are all within a few seconds of the actual creation time (I assume), except for disks or instances where I think we can reattach or relaunch them and "restart the clock".  Then these resources will not be cleaned up, but only if they are continually being reattached or relaunched .... in which case they are in use, and thus I think we don't _want_ to clean them up.

cjwagner: Ok that makes sense, but are such resources leaked permanently or will they later change to a state where we will clean them up? This sounds good as long as the resources will eventually be recovered and aren't permanently leaked.
Sorry, I'm not familiar with this code.

justinsb: They'll be deleted after the TTL (4 hours, IIRC).  They only don't get cleaned up if we have jobs that keep deleting and recreating them, but that's not really leaking.

cjwagner (approved):

PR #134, opened 2022-07-08
Distinguish between resources busy and unknown resource type

### LINKED ISSUE #126: Boskos client does not distinguish between incorrect resource type and no resources available
Currently, when acquiring a resource using AcquireWait(), the client makes no distinction between a resource type that does not exist (which could arise from a user typo) vs. a situation where all resources are busy. 

Unfortunately, the Boskos Server sends the 404 status code in both situation - [see code here](https://github.com/kubernetes-sigs/boskos/blob/9fae38623263fd9251734b5dd5d78a0f09c3876a/handlers/handlers.go#L71). The Boskos client then looks at the status code [here](https://github.com/kubernetes-sigs/boskos/blob/d007e4465eb6fcc4ab275698353fe3bafbeb28d5/client/client.go#L191).

One easy and backwards-compatible fix would be to add a new option to the client that, when set, would distinguish between the two situation based on the text returned from the HTTP call. The return text for the two different situations is:

- Acquire failed: resource type "my-resource-type" does not exist
- Acquire failed: no available resource my-resource-type, try again later

The current problem is that if the user accidentally asks for a resource type that does not exist, they will be stuck in a loop forever with no hope of unblocking except for a Context timeout, which, imo, is not an acceptable situation.

bartekmi:
Closes #126

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -43,11 +43,6 @@ import (
 	"sigs.k8s.io/boskos/storage"
 )
 
-const (
-	// resourceTypeErrorSnippet is present in the HTTP response if user requests a resource type that does not exist.
-	resourceTypeErrorSnippet = "resource type"
-)
-
 var (
 	// ErrNotFound is returned by Acquire() when no resources are available.
 	ErrNotFound = errors.New("resources not found")
@@ -432,7 +427,9 @@ func (c *Client) acquire(rtype, state, dest, requestID string) (*common.Resource
 			// resource type is to check the text of the accompanying error message.
 			if c.DistinguishNotFoundVsTypeNotFound {
 				if bytes, err := io.ReadAll(resp.Body); err == nil {
-					if strings.Contains(string(bytes), resourceTypeErrorSnippet) {
+					cutset := " \n\t"
+					str := strings.Trim(string(bytes), cutset)
+					if str == strings.Trim(common.ResourceTypeNotFoundMessage(rtype), cutset) {
 						return false, ErrTypeNotFound
 					}
 				}
--- a/client/client.go
+++ b/client/client.go
@@ -427,9 +427,8 @@ func (c *Client) acquire(rtype, state, dest, requestID string) (*common.Resource
 			// resource type is to check the text of the accompanying error message.
 			if c.DistinguishNotFoundVsTypeNotFound {
 				if bytes, err := io.ReadAll(resp.Body); err == nil {
-					cutset := " \n\t"
-					str := strings.Trim(string(bytes), cutset)
-					if str == strings.Trim(common.ResourceTypeNotFoundMessage(rtype), cutset) {
+					errorMsg := string(bytes)
+					if strings.Contains(errorMsg, common.ResourceTypeNotFoundMessage(rtype)) {
 						return false, ErrTypeNotFound
 					}
 				}
--- a/client/client.go
+++ b/client/client.go
@@ -43,6 +43,11 @@ import (
 	"sigs.k8s.io/boskos/storage"
 )
 
+const (
+	// resourceTypeErrorSnippet is present in the HTTP response if user requests a resource type that does not exist.
+	resourceTypeErrorSnippet = "resource type"
+)
+
 var (
 	// ErrNotFound is returned by Acquire() when no resources are available.
 	ErrNotFound = errors.New("resources not found")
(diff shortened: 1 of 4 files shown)


stevekuznetsov (commented):
Why would it not be appropriate to do this at the server level?

client/client.go
```diff
@@ -43,6 +43,11 @@ import (
 	"sigs.k8s.io/boskos/storage"
 )
 
+const (
+	// resourceTypeErrorSnippet is present in the HTTP response if user requests a resource type that does not exist.
+	resourceTypeErrorSnippet = "resource type"
```
chaodaiG, line 48: this is a pretty loose contract imo. The error message is "fmt.Sprintf("resource type %q does not exist", r.rType)" according to https://github.com/kubernetes-sigs/boskos/blob/9fae38623263fd9251734b5dd5d78a0f09c3876a/ranch/ranch.go#L62. A couple of thoughts here:

- Option 1: since the rType is known by the time the condition is checked, I think we can do strict full string matching.
- Option 2: add a function directly in `rancher`, side by side with `func (r ResourceTypeNotFound) Error()` function, called something like `func isResourceTypeNotFoundErrorString`, and add a unit test to ensure that these two functions always agree with each other

client/client.go
```diff
@@ -51,13 +56,20 @@ var (
...
 
 // Client defines the public Boskos client object
 type Client struct {
 	// Dialer is the net.Dialer used to establish connections to the remote
 	// boskos endpoint.
 	Dialer DialerWithRetry
+	// DistinguishNotFoundVsTypeNotFound, if set, will make it possible to distinguish between
+	// ErrNotFound and ErrTypeNotFound. For backwards-compatibility, this flag is off by
+	// default.
+	DistinguishNotFoundVsTypeNotFound bool
```
chaodaiG, line 72: do we need to define a flag somewhere to enable this?

client/client.go
```diff
@@ -432,7 +427,9 @@ func (c *Client) acquire(rtype, state, dest, requestID string) (*common.Resource
 			// resource type is to check the text of the accompanying error message.
 			if c.DistinguishNotFoundVsTypeNotFound {
 				if bytes, err := io.ReadAll(resp.Body); err == nil {
-					if strings.Contains(string(bytes), resourceTypeErrorSnippet) {
+					cutset := " \n\t"
```
chaodaiG, line 430: is this equivalent to https://pkg.go.dev/strings#TrimSpace?

client/client.go
bartekmi: I've implemented something close to Option 2 where there is only "once source of truth" for the message shared by the client and server - defined in common.go

client/client.go
bartekmi: Since, by default, we preserve the old behavior, I don't believes that the extra setting via flags is necessary. From the user's point of view, their code would look like:

```
client := boskos.NewClient(...)           # as before
client.DistinguishNotFoundVsTypeNotFound = true
```

bartekmi:
> Why would it not be appropriate to do this at the server level?

Having implemented @[chaodaiG](https://github.com/chaodaiG)'s suggestion, I believe the current implementation should be closer to what you envisioned - one source of truth for client and server.

By not actually depending on modified functionality of the server, user's don't need to upgrade their server to take advantage of the new functionality.

client/client_test.go
```diff
@@ -60,18 +60,18 @@ func TestAcquire(t *testing.T) {
 	}{
 		{
 			name:      "request error",
-			serverErr: "Acquire failed: no available resource my-resource-type, try again later",
```
chaodaiG: isn't `Acquire failed: ` part of the error message from the server?

client/client.go
chaodaiG: ack

client/client_test.go
bartekmi: Yeah - that's a good catch - I switch from an exact string match to doing a substring compare because the message returned from the server is actually a composite of the straight error message and the constant string "Acquire failed:"

chaodaiG (approved):

stevekuznetsov:
> By not actually depending on modified functionality of the server, user's don't need to upgrade their server to take advantage of the new functionality.

Shouldn't users be updating their `boskos` installation regularly anyway? The bumper utility automates it, for all the deployments I know of, anyway.

Seems like a more durable solution (folllow-up would be fine) to depend on some structured data instead of a happenstance human-readable string would be good. Allowing new clients to figure it out for old servers seems useful though, I guess.

chaodaiG:
> Shouldn't users be updating their `boskos` installation regularly anyway? The bumper utility automates it, for all the deployments I know of, anyway.

Boskos is deployed in "build clusters", in our use case the automation of most of the build clusters are beyond our control, so we don't really have guarantee that they are managed by the bumper utility, or they are up-to-date.

> Seems like a more durable solution (folllow-up would be fine) to depend on some structured data instead of a happenstance human-readable string would be good.

This is something I have thought about as well, but didn't have a good solution in this case. Now that you mentioned, may be something like json string like `{"reason": "resource type not found", "type": "<RESOURCE_TYPE>"}` would work. But this would not work with old server as you mentioned.

> Allowing new clients to figure it out for old servers seems useful though, I guess.

The current implementation doesn't change the error string from the server, so it already supports it.

PR #135, opened 2022-07-09
Don't delete AWS IAM roles that have been created too recently

olemarkus:
kOps e2e tests using IAM roles assumed by Pods have been failing randomly due to IAM roles mysteriously disappearing. Since it may take 5-10 minutes between an IAM role being created and the pods consuming a role, there is a window where the AWS janitor can delete unused roles. This PR adds an additional check that the role has not been created too recently before it is removed.


olemarkus:
/close 
Superseded by https://github.com/kubernetes-sigs/boskos/pull/133

PR #136, opened 2022-07-27
Support target groups deletion in aws-janitor repo.

youyangl:
For now, the target groups(https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#TargetGroup) are not GCed by the aws-janitor, this change intends to support that.

TESTED=I run make, make test and make verify locally, not error found.
Tried:
go run cmd/aws-janitor/main.go --dry-run=true --all=true
go run cmd/aws-janitor/main.go --dry-run=true --all=true --enable-target-group-clean=true
locally
The log looks good to me

### DIFF
--- a/aws-janitor/resources/list.go
+++ b/aws-janitor/resources/list.go
@@ -39,6 +39,9 @@ type Options struct {
 
 	// Whether to actually delete resources, or just report what would be deleted.
 	DryRun bool
+
+	// If true, clean target groups.
+	EnableTargetGroupClean bool
 }
 
 type Type interface {
@@ -78,6 +81,8 @@ var RegionalTypeList = []Type{
 	Addresses{},
 	ElasticFileSystems{},
 	SQSQueues{},
+	// ELBV2 target groups,
+	TargetGroups{},
 }
 
 // Non-regional AWS resource types, in dependency order
--- a/aws-janitor/resources/target_groups.go
+++ b/aws-janitor/resources/target_groups.go
@@ -1,0 +1,140 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package resources
+
+import (
+	"fmt"
+	"time"
+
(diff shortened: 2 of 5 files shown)


aws-janitor/resources/target_groups.go
```diff
@@ -0,0 +1,133 @@
+/*
```
youyangl, line 1: It's copied from [elbv2.go](https://github.com/kubernetes-sigs/boskos/blob/3416530a0ed99825d6a74764ad57e60f98d5e0e7/aws-janitor/resources/elbv2.go), but I'm not 100% sure whether some logics for load balancer cleanup should also be implemented for target groups

aws-janitor/resources/target_groups.go
```diff
@@ -0,0 +1,133 @@
...
+			tgTags[a.ARN()] = nil
+		}
+		return true
+	}
+
+	if err := svc.DescribeTargetGroupsPages(&elbv2.DescribeTargetGroupsInput{}, pageFunc); err != nil {
+		return err
+	}
+
+	fetchTagErr := incrementalFetchTags(tgTags, 20, func(tgArns []*string) error {
```
youyangl, line 55: I'm not sure whether do we need to use incrementalFetchTags here, maybe we can just use fromEC2Tags instead?

chaodaiG (changes requested):
This change is larger than I thought. Could you make it configurable, so that it won't be turned on unless an arg is passed in? Or alternatively not fail this binary when the step failed, log an error instead

youyangl:
> 

I think we can add a new field: EnableTargetGroupClean in resource options and set its value using flags passed to binary.
If EnableTargetGroupClean is false, just return nil

chaodaiG (approved):
/hold
Contingent approve based on your team's review on the business logic

aws-janitor/resources/target_groups.go
chemikadze: TargetGroups don't seem to have Tags on them, so we can't use fromEC2Tags here, and ELBv2's DescribeTags is limited to 20 ARNs per request, so we do need to use incrementalFetchTags here.

chemikadze (approved):

aws-janitor/resources/list.go
```diff
@@ -39,6 +39,9 @@ type Options struct {
 
 	// Whether to actually delete resources, or just report what would be deleted.
 	DryRun bool
+
+	// If true, clean target groups.
+	EnableTargetGroupClean bool
```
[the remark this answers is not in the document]
chaodaiG: feel free to remove this feature flag once tested and confirmed it working in prod. Would be good to announce this breaking change on Slack and even better on sig-testing bi-weekly meeting

chaodaiG:
fyi @youyangl , the image push job was fixed by @olemarkus : https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-boskos-push-images/1553081733878910977 , the changes from this PR should also be included in the images

PR #140, opened 2022-07-29
Bump alpine version used for the boskosctl image

olemarkus:
Build still breaking. Old alpine version this time.

/cc @alvaroaleman @chaodaiG

### DIFF
--- a/images/boskosctl/Dockerfile
+++ b/images/boskosctl/Dockerfile
@@ -15,7 +15,7 @@
 # Installs a few extra tools folks might want to use when running boskosctl.
 
 ARG go_version
-ARG alpine_version=3.14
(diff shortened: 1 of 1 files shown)


chaodaiG:
Thank you for digging into this, I'm still very confused why the image build job still failed after the revert the test-infra deps bump. If this still doesn't work we'll probably need to look deeper into the root cause before another attempt

/lgtm
/approve

olemarkus:
I am guessing not everything was completely reverted. Not sure. But the symptoms are pretty consistent with outdated golang, and that fixed the first error (go.mod weirdness). This fixed an image pull issue as there were no golang 1.18-alpine3.14 image.

Issue #142, opened 2022-08-09
gcp_janitor.py isn't thread-safe: multiple threaded invocations can corrupt GCloud config file


pigboysid:
In one of my Boskos instances, I've observed failures when invoking [gcp_janitor.py](https://github.com/kubernetes-sigs/boskos/blob/master/cmd/janitor/gcp_janitor.py) of the following form:

```
failed to clean up project asm-boskos-shared-vpc-svc-188, error info: ERROR: gcloud failed to load: Source contains parsing errors: '/root/.config/gcloud/configurations/config_default'
	[line 13]: 'oogleapis.com/\n'
    parsed_config.read(properties_path)
    self._read(fp, filename)
    raise e
	[line 13]: 'oogleapis.com/\n'
```

My analysis of [gcp_janitor.py](https://github.com/kubernetes-sigs/boskos/blob/master/cmd/janitor/gcp_janitor.py) makes me believe it's not thread-safe, specifically when running `gcloud config set`.

The first place we run that is in [line 511](https://github.com/kubernetes-sigs/boskos/blob/master/cmd/janitor/gcp_janitor.py#L511), where we run:

```
gcloud config set billing/quota-project <xyz>
```

The second is in [line 588](https://github.com/kubernetes-sigs/boskos/blob/master/cmd/janitor/gcp_janitor.py#L588), where we run:

```
gcloud config set api_endpoint_overrides/gkehub https://<gkehub-url>/
```

The janitor itself, [janitor.go](https://github.com/kubernetes-sigs/boskos/blob/master/cmd/janitor/janitor.go), invokes `gcp_janitor.py` inside Goroutines, which run in parallel; I believe that if multiple threads attempt to run `gcloud config set` simultaneously, it can corrupt the GCloud config file (in my case, `/root/.config/gcloud/configurations/config_default`), which is shared among all threads. This renders any future attempts to run `gcp_janitor.py` futile, because the GCloud config file is irrevocably corrupted.

I believe I have a fix for this, involving setting `os.environ` rather than running `gcloud config set`. Specifically, you can replace the commands with environment variables like this:

```
gcloud config set billing/quota-project <xyz>
-> CLOUDSDK_BILLING_QUOTA_PROJECT=<xyz>

gcloud config set api_endpoint_overrides/gkehub https://<gkehub-url>/
-> CLOUDSDK_API_ENDPOINT_OVERRIDES_GKEHUB=https://<gkehub-url>
```

[...]

PR #144, opened 2022-08-09
Make GCP janitor thread-safe

pigboysid:
As discussed in https://github.com/kubernetes-sigs/boskos/issues/142,
the GCP janitor isn't thread-safe: sometimes, it can corrupt the GCloud
config file (which is on-disk, and which is shared by all threads).

Instead of using "gcloud config set X/Y Z", use
"os.environ['CLOUDSDK_X_Y'] = Z".

### DIFF
--- a/cmd/janitor/gcp_janitor.py
+++ b/cmd/janitor/gcp_janitor.py
@@ -508,17 +508,7 @@ def activate_service_account(service_account):
 
 def set_quota_project(project):
     print('[=== Setting quota_project %s ===]' % project)
-    cmd = [
-        'gcloud', 'config', 'set', 'billing/quota_project',
-        '%s' % project,
-    ]
-    log('running %s' % cmd)
-
-    try:
-        subprocess.check_call(cmd)
-    except subprocess.CalledProcessError:
-        print('Error try to set quota_project: %s' % project, file=sys.stderr)
-        return 1
+    os.environ['CLOUDSDK_BILLING_QUOTA_PROJECT'] = '%s' % project
     return 0
 
 # Returns whether the specified GCP API is enabled on the provided project.
@@ -585,11 +575,7 @@ def main(project, days, hours, filt, rate_limit, service_account, additional_zon
         if not api_enabled(project, api):
             continue
         if api in gkehub_apis:
-            cmd = "gcloud config set api_endpoint_overrides/gkehub https://{}/".format(api)
-            try:
-                subprocess.run(cmd, shell=True, text=True, check=True)
-            except (subprocess.CalledProcessError, ValueError) as exc:
-                log('Cannot set GKE HUB endpoint %s with %r, continue' % (api, exc))
+            os.environ['CLOUDSDK_API_ENDPOINT_OVERRIDES_GKEHUB'] = 'https://{}/'.format(api)
         for res in resources:
             log('Try to search for %r with condition %r, managed %r' % (
                 res.name, res.condition, res.managed))


pigboysid:
In case it expedites the approval of this review, I'd like to say:

You'll notice that I've removed the checks which validate that the `gcloud config set` commands run without error. My justifications for this are as follows:

- I'm now setting an environment variable rather than running a command, so there's no longer an error to check.
- I tested `gcloud config set billing/quota_project <xyz>` on the command line myself. It appears to do no validation of the project argument `<xyz>` at all: even if that project doesn't exist, it allows you to set it. Thus, it seems to me that the error condition (i.e., `gcloud config set` fails) would never have been triggered.
- Likewise, I tested `gcloud config set api_endpoint_overrides/gkehub`. It _does_ emit an error if the argument isn't of the form `http://<some-url>/` or `https://<some-url>/`, but otherwise, it appears to do no URL validation at all (i.e., it doesn't attempt to connect to the URL you provide). As above, it seems to me that the error condition in the old code would never have been triggered.

cmd/janitor/gcp_janitor.py
```diff
@@ -508,17 +508,7 @@ def activate_service_account(service_account):
...
-        '%s' % project,
-    ]
-    log('running %s' % cmd)
-
-    try:
-        subprocess.check_call(cmd)
-    except subprocess.CalledProcessError:
-        print('Error try to set quota_project: %s' % project, file=sys.stderr)
-        return 1
+    os.environ['CLOUDSDK_BILLING_QUOTA_PROJECT'] = '%s' % project
```
listx, line 511: My Python is rusty so I stackoverflowed and I found https://stackoverflow.com/questions/5971312/how-to-set-environment-variables-in-python#comment61988654_5971326 --- do the concerns there about child processes vs parent process apply here as well? I assume "no" and this is fine, but just wanted to ask.

pigboysid: Hey Linus: thanks for the prompt attention. To ensure I'm responding to your actual question, you're referring to this comment, yes?

> Caution: to quote from @Evan's reference above, Such changes to the environment affect subprocesses started with os.system(), popen() or fork() and execv(). In other words, keep in mind that this approach won't modify the way your program is running, only the way your program's children run.

In this case, `gcp_janitor.py` *is* the child process (it's spawned from `janitor.go`). So, settings that we make in `gcp_janitor.py` only affect its own specific invocation, not others spawned by `janitor.go`.

You can see that overriding the environment variable *does* affect subsequent `subprocess` invocations in the same script. Some sample code:
```python
import os, subprocess

GCLOUD_CMD = "gcloud config get api_endpoint_overrides/gkehub"

def set_gkehub(api):
  os.environ["CLOUDSDK_API_ENDPOINT_OVERRIDES_GKEHUB"] = f'https://{api}/'

def run_cmd(pstr):
  print(f"-- {pstr}")
  subprocess.run(GCLOUD_CMD, shell=True, text=True, check=True)

run_cmd("No env override")

gkehub_apis = {'gkehub.googleapis.com', 'staging-gkehub.sandbox.googleapis.com', 'autopush-gkehub.sandbox.googleapis.com'}
for api in gkehub_apis:
  set_gkehub(api)
  run_cmd(f"Env override with {api}")
```
And here's the output when I run that code:
```
-- No env override
(unset)
-- Env override with autopush-gkehub.sandbox.googleapis.com
https://autopush-gkehub.sandbox.googleapis.com/
-- Env override with staging-gkehub.sandbox.googleapis.com
https://staging-gkehub.sandbox.googleapis.com/
-- Env override with gkehub.googleapis.com
https://gkehub.googleapis.com/
```
The bottom line is: `os.environ` sets the environment only for a given script and its children, and separate invocations of `gce_janitor.py` (which don't affect each others' environments) are spawned by `janitor.go`.

pigboysid: Sorry, Linus, I should add one clarification:

Calls to `subprocess` use `popen` under the hood. That's why setting `os.environ` inside my sample code above does affect subsequent calls to `subprocess`.

listx: > To ensure I'm responding to your actual question, you're referring to this comment, yes?

Yes. (Sorry I didn't make it clear in my original question.)

LGTM

pigboysid: Thank you, @listx!

By the way, I see a label on this PR saying `needs-ok-to-test`. Are you able to authorize that?

PR #145, opened 2022-08-10
Delete outdated key pairs in aws-janitor repo

youyangl:
Run make, make build, make test, make verify locally, no error found


chaodaiG:
/ok-to-test
please have someone who has context approving this PR before assigning to me.
/unassign

aws-janitor/resources/key_pairs.go
```diff
@@ -0,0 +1,93 @@
...
+
+	return set, nil
+}
+
+type keyPair struct {
+	KeyFingerprint string
+}
+
+func (kp keyPair) ARN() string {
+	return kp.KeyFingerprint
```
chemikadze, line 88: From what I understand we need globally unique identifier here, including region, as keypairs are regional resource. Also, same keypair with same fingerprint could be imported multiple times, so we probably need to use name as part of synthetic ARN rather than fingerprint.

See addresses.go for example.

go.mod
```diff
@@ -11,23 +11,23 @@ require (
 	github.com/IBM-Cloud/power-go-client v1.1.11
 	github.com/IBM/go-sdk-core/v5 v5.10.2
 	github.com/IBM/platform-services-go-sdk v0.28.2
-	github.com/aws/aws-sdk-go v1.37.22
```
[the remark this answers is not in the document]
youyangl: The old version github.com/aws/aws-sdk-go/service/ec2 lib doesn't have field: CreateTime in KeyPairInfo. If we don't use the new AWS SDK here, we can't delete key pairs using timestamp.

aws-janitor/resources/key_pairs.go
youyangl: I see, then I think we can compose the ARN using account , region, key pair id.

aws-janitor/resources/key_pairs.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+type keyPair struct {
+	Account string
+	Region  string
+	ID      string
+}
+
+func (kp keyPair) ARN() string {
+	return fmt.Sprintf("arn:aws:ec2:%s:%s:keypair/%s", kp.Region, kp.Account, kp.ID)
```
chemikadze, line 97: Nit: is this valid ARN, or should we leave a comment that it's synthetic?

aws-janitor/resources/list.go
```diff
@@ -83,6 +83,7 @@ var RegionalTypeList = []Type{
 	SQSQueues{},
 	// ELBV2 target groups,
 	TargetGroups{},
+	KeyPairs{},
```
chemikadze, line 86: nit: do we need a feature flag like for target groups?

chemikadze (approved):

chaodaiG (approved):

PR #149, opened 2022-09-13
drop require directives

dprotaso:
I don't think they're needed anymore


chaodaiG:
/ok-to-test
I'm not sure the assumption holds true

chaodaiG:
seems like you're right, so
/lgtm
/approve

PR #152, opened 2022-10-10
Add support for cleaning IBM Cloud VPC resources

Amulyam24:
This PR adds supports for cleaning up IBM Cloud VPC resources.

- The VPC resources are cleaned up in a given **_region_** belonging to a particular **_resource group_**, both of which are passed in the resource user data
- Cleanup added for VSIs, VPCs, subnets, gateways and floating IPs.


Prajyot-Parab (approved):
/lgtm

internal/ibmcloud-janitor/resources/vpc.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	for _, vpc := range vpcList.Vpcs {
+		_, err = client.DeleteVPC(&vpcv1.DeleteVPCOptions{
+			ID: vpc.ID,
+		})
+		if err != nil {
+			return errors.Wrapf(err, "failed to delete the VPC %q", *vpc.Name)
+		}
+	}
```
mkumatag, line 50: This seems like we are deleting all the vpc in that given region, I would like to understand the overall design for the vpc resource, it will be great if you can put together some doc for the same(may updating the readme file will be helpful)

internal/ibmcloud-janitor/resources/vpc_networks.go
```diff
@@ -0,0 +1,87 @@
...
+			}
+			resourceLogger.WithFields(logrus.Fields{"name": pg.Name}).Info("Successfully deleted the gateway")
+		}
+		_, err = client.DeleteSubnet(&vpcv1.DeleteSubnetOptions{ID: subnet.ID})
+		if err != nil {
+			return errors.Wrapf(err, "failed to delete the subnet %q", *subnet.Name)
+		}
+	}
+
+	fips, _, err := client.ListFloatingIps(&vpcv1.ListFloatingIpsOptions{
```
mkumatag, line 69: some additional comment about why we are doing this will be helpful.

internal/ibmcloud-janitor/resources/vpc_networks.go
```diff
@@ -0,0 +1,87 @@
...
+import (
+	"github.com/IBM/vpc-go-sdk/vpcv1"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+type VpcNetwork struct{}
+
+// Cleans up the FIPs, subnets and its attached gateways in a given region
+func (VpcNetwork) cleanup(options *CleanupOptions) error {
```
mkumatag, line 28: and also it will be helpful to add more comments with mentioning step1, step2, step3 etc.. e.g: step1: delete publicgateway, step2: floating point ip etc..

mkumatag: you also might need to handle the load balancers

common/ibmcloud/common.go
```diff
@@ -29,40 +29,60 @@ const (
...
 	Region            = "region"
 	Zone              = "zone"
+	ResourceGroup     = "resource-group"
 )
 
-type ResourceData struct {
+type PowervsResourceData struct {
 	ServiceInstanceID string
-	Region            string
 	Zone              string
```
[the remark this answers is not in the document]
Amulyam24: We are adding resource group as a part of user data for PowerVS but it is not being loaded here as the janitor doesn't need it.

internal/ibmcloud-janitor/resources/vpc_networks.go
Amulyam24: I'm planning to handle the load balancer cleanup in a follow up PR(since we are using the standard cluster template in the CAPI e2e tests as of now). Does it sound okay?

Amulyam24:
Changes applied as per suggestions, PTAL

mkumatag (approved):
/lgtm

chaodaiG (approved):

PR #150, opened 2022-09-15
Add youyangl to the aws-janitor-contributors alias

youyangl:
Add myself to the aws-janitor-contributors alias
To fix some aws-janitor releated issues, I sent some PRs:
https://github.com/kubernetes-sigs/boskos/pull/136
https://github.com/kubernetes-sigs/boskos/pull/145
https://github.com/kubernetes-sigs/boskos/pull/146
to the aws-janitor package. To make further reviews more convenient, i intend to add myself to the approver list.

PR #153, opened 2022-10-18
Clean expired DNS zones when deleting resource records.

youyangl:


elton-d (approved):

chaodaiG:
/ok-to-test
Per discussion from https://github.com/kubernetes-sigs/boskos/pull/150 that there are existing reviewers for aws-janitor who have much better contexts than I do, so
/unassign
Assign back to me when you get initial review from the existing owners

chaodaiG (approved):

PR #154, opened 2022-11-15
Delete S3 bucket in aws janitor

youyangl:
This change deletes expired s3 buckets in aws janitor

### DIFF
--- a/aws-janitor/resources/s3_bucket.go
+++ b/aws-janitor/resources/s3_bucket.go
@@ -1,0 +1,143 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package resources
+
+import (
+	"context"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go/service/s3/s3manager"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+type S3Bucket struct{}
+
+// IsManagedS3Bucket checks whether the bucket should be managed by janitor.
+func IsManagedS3Bucket(opts Options, region string, bucketName string) (bool, error) {
+	svc := s3.New(opts.Session, &aws.Config{Region: aws.String(region)})
+	respTags, err := svc.GetBucketTagging(&s3.GetBucketTaggingInput{Bucket: aws.String(bucketName)})
+	if err != nil {
+		return false, err
+	}
+	tags := fromS3Tags(respTags.TagSet)
+	return opts.ManagedPerTags(tags), nil
+}
+
+func (S3Bucket) MarkAndSweep(opts Options, set *Set) error {
+	logger := logrus.WithField("options", opts)
+	if !opts.EnableS3BucketsClean {
+		logger.Info("S3 bucket cleanup is not enabled")
+		return nil
+	}
+	svc := s3.New(opts.Session, &aws.Config{Region: aws.String(opts.Region)})
+
+	// ListBucket lists all buckets owned by the authenticated sender of the request
+	// https://pkg.go.dev/github.com/aws/aws-sdk-go/service/s3#S3.ListBuckets
+	// regardless the region configured in client.
+	resp, err := svc.ListBuckets(&s3.ListBucketsInput{})
+	if err != nil {
+		return err
+	}
+
+	for _, b := range resp.Buckets {
+		bucketName := *b.Name
+
+		region, err := s3manager.GetBucketRegion(context.Background(), opts.Session, bucketName, opts.Region)
+		if err != nil || region != opts.Region {
+			continue
+		}
+		respTags, err := svc.GetBucketTagging(&s3.GetBucketTaggingInput{Bucket: aws.String(bucketName)})
+		if err != nil {
+			continue
+		}
+		tags := fromS3Tags(respTags.TagSet)
+		if !set.Mark(opts, s3Bucket{bucketName: bucketName}, nil, tags) {
+			continue
+		}
+		var objectsToDelete []*string
+		pageFunc := func(page *s3.ListObjectsV2Output, _ bool) bool {
+			for _, object := range page.Contents {
+				logger.Warningf("Deleting object: %s at bucket: %s", *object.Key, bucketName)
+				objectsToDelete = append(objectsToDelete, object.Key)
+			}
+			return true
+		}
+		// Before deleting the bucket, delete its objects.
+		if err := svc.ListObjectsV2Pages(&s3.ListObjectsV2Input{Bucket: aws.String(bucketName)}, pageFunc); err != nil {
+			logger.Warningf("ListObjectsV2Pages failed for bucket: %s : %v", bucketName, err)
+			continue
+		}
+
+		logger.Warningf("Deleting bucket: %s", bucketName)
+		if opts.DryRun {
+			continue
+		}
+		if len(objectsToDelete) > 0 {
+			for _, objectKey := range objectsToDelete {
+				if _, err := svc.DeleteObject(&s3.DeleteObjectInput{Bucket: &bucketName, Key: objectKey}); err != nil {
+					logger.Warningf("DeleteObject failed for bucket: %s, object: %s: %v", bucketName, *objectKey, err)
+				}
+			}
+		}
+		if _, err := svc.DeleteBucket(&s3.DeleteBucketInput{Bucket: &bucketName}); err != nil {
+			logger.Warningf("DeleteBucket failed for bucket: %s : %v", bucketName, err)
+		}
+	}
+	return nil
+}
+
+func (S3Bucket) ListAll(opts Options) (*Set, error) {
+	set := NewSet(0)
+	if !opts.EnableS3BucketsClean {
+		return set, nil
+	}
+	svc := s3.New(opts.Session, &aws.Config{Region: aws.String(opts.Region)})
+
+	// ListBucket lists all buckets owned by the authenticated sender of the request
+	// https://pkg.go.dev/github.com/aws/aws-sdk-go/service/s3#S3.ListBuckets
+	// regardless the region configured in client.
+	resp, err := svc.ListBuckets(&s3.ListBucketsInput{})
+	if err != nil {
+		return set, err
+	}
+	for _, b := range resp.Buckets {
+		bucketName := *b.Name
+		region, err := s3manager.GetBucketRegion(context.Background(), opts.Session, bucketName, opts.Region)
+		if err != nil || region != opts.Region {
+			continue
+		}
+		now := time.Now()
+		set.firstSeen[bucketName] = now
+	}
+
+	return set, errors.Wrapf(err, "couldn't list buckets for %q in %q", opts.Account, opts.Region)
+}
+
+type s3Bucket struct {
+	bucketName string
+}
+
+func (s s3Bucket) ARN() string {
+	return s.bucketName
+}
+
+func (s s3Bucket) ResourceKey() string {
+	return s.ARN()
+}
--- a/cmd/aws-janitor-boskos/main.go
+++ b/cmd/aws-janitor-boskos/main.go
@@ -59,6 +59,7 @@ var (
 	enableVPCEndpointsClean    = flag.Bool("enable-vpc-endpoints-clean", false, "If true, clean vpc endpoints.")
 	skipRoute53ManagementCheck = flag.Bool("skip-route53-management-check", false, "If true, skip managed zone check and managed resource name check.")
 	enableDNSZoneClean         = flag.Bool("enable-dns-zone-clean", false, "If true, clean DNS zones.")
+	enableS3BucketsClean       = flag.Bool("enable-s3-buckets-clean", false, "If true, clean S3 buckets.")
 
 	excludeTags common.CommaSeparatedStrings
 	includeTags common.CommaSeparatedStrings
@@ -196,6 +197,7 @@ func cleanResource(res *common.Resource) error {
 		EnableVPCEndpointsClean:    *enableVPCEndpointsClean,
 		SkipRoute53ManagementCheck: *skipRoute53ManagementCheck,
 		EnableDNSZoneClean:         *enableDNSZoneClean,
+		EnableS3BucketsClean:       *enableS3BucketsClean,
 	}
 
 	logrus.WithField("name", res.Name).Info("beginning cleaning")
--- a/cmd/aws-janitor/main.go
+++ b/cmd/aws-janitor/main.go
@@ -42,7 +42,7 @@ import (
 var (
 	maxTTL                     = flag.Duration("ttl", 24*time.Hour, "Maximum time before attempting to delete a resource. Set to 0s to nuke all non-default resources.")
 	region                     = flag.String("region", "", "The region to clean (otherwise defaults to all regions)")
-	path                       = flag.String("path", "", "S3 path for mark data (required when -all=false)")
+	path                       = flag.String("path", "", "S3 path for mark data, the bucket must be marked with excludeTags if s3 bucket clean is on (required when -all=false)")
 	cleanAll                   = flag.Bool("all", false, "Clean all resources (ignores -path)")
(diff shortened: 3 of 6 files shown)


aws-janitor/resources/s3_bucket.go
```diff
@@ -0,0 +1,97 @@
...
+type S3Bucket struct{}
+
+func (S3Bucket) MarkAndSweep(opts Options, set *Set) error {
+	logger := logrus.WithField("options", opts)
+	if !opts.EnableS3BucketClean {
+		logger.Info("Disable s3 bucket clean")
+		return nil
+	}
+	svc := s3.New(opts.Session, &aws.Config{Region: aws.String(opts.Region)})
+	resp, err := svc.ListBuckets(&s3.ListBucketsInput{})
```
youyangl, line 38: The ListBucket lists all buckets owned by the authenticated sender of the request([https://github.com/aws/aws-sdk-go/blob/v1.44.137/service/s3/api.go#L6158]), which means that even though we config the region when initializing the client(for example, "us-west-1"), the ListBuckets still returns buckets in all regions, not limited to "us-west-1".
Thus, at line 44, we need to drop buckets which are not in opts.Region, otherwise, the GetBucketTagging will fail since this method only works for buckets with the same region as what we config in client.

aws-janitor/resources/s3_bucket.go
```diff
@@ -0,0 +1,97 @@
...
+	if err != nil {
+		return err
+	}
+	for _, b := range resp.Buckets {
+		bucketName := *b.Name
+		if !strings.Contains(bucketName, opts.Region){
```
youyangl, line 44: The easiest way to check the region of bucket is to extract the region from the bucket name.
However, not all buckets contain the region in bucket name, we should find another way to get the bucket region.

aws-janitor/resources/s3_bucket.go
```diff
@@ -0,0 +1,97 @@
...
+			continue
+		}
+		respTags, err := svc.GetBucketTagging(&s3.GetBucketTaggingInput{Bucket: aws.String(bucketName)})
+		if err != nil {
+			continue
+		}
+		tags := fromS3Tags(respTags.TagSet)
+		if !set.Mark(opts, s3Bucket{bucketName: bucketName}, nil, tags) {
```
youyangl, line 52: For now, we only delete expired s3 buckets, do we also need to delete empty bucket(bucket without any object)?

cmd/aws-janitor/main.go
```diff
@@ -191,8 +193,24 @@ func markAndSweep(opts resources.Options, region string) error {
...
 	sweepCount = res.MarkComplete()
-	if err := res.Save(opts.Session, s3p); err != nil {
-		return errors.Wrapf(err, "Error saving %q", *path)
+
+	saveSet := true
+	if *enableS3BucketClean{
+		if !s3path.CheckBucketExist(opts.Session, opts.Region, s3p.Bucket) {
+			logrus.Warningf("DeleteSet for bucket: %s", s3p.Bucket)
+			if !*dryRun {
+				if err := res.DeleteSet(opts.Session, s3p); err != nil {
```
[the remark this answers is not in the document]
chemikadze: Taking that bucket must be empty to be removed, this should never happen?

aws-janitor/resources/s3_bucket.go
```diff
@@ -0,0 +1,97 @@
...
+			continue
+		}
+		if !opts.DryRun {
+			if _, err := svc.DeleteBucket(&s3.DeleteBucketInput{Bucket: &bucketName}); err != nil {
```
chemikadze, line 56: It looks like only empty buckets can be deleted, do we need to delete objects first? https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteBucket.html

cmd/aws-janitor/main.go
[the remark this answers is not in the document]
chemikadze: Also, should we explicitly skip state bucket instead during cleanup? Otherwise janitor will loose metadata about other objects.

aws-janitor/resources/set.go
```diff
@@ -89,11 +94,16 @@ func (s *Set) Save(sess *session.Session, p *s3path.Path) error {
...
+	// Field Tagging must be encoded as URL Query parameters, (For example, "Key1=Value1")
+	// https://docs.aws.amazon.com/sdk-for-go/api/service/s3/#PutObjectInput
+	stateTag := fmt.Sprintf("%s=true", JanitorStateTag)
+
 	_, err = svc.PutObject(&s3.PutObjectInput{
 		Bucket:       aws.String(p.Bucket),
 		Key:          aws.String(p.Key),
 		Body:         bytes.NewReader(b),
 		CacheControl: aws.String("max-age=0"),
+		Tagging:      &stateTag,
```
youyangl, line 106: For now, it's really hard to check whether the object is a janitor state object or not since its key is derived from the "path" flag 
https://github.com/youyangl/boskos/blob/8ed406269010654ac4a50e70e62abf17864a7ef5/cmd/aws-janitor/main.go#L45. which can be customized by user.
One possible solution is that we can set the "Tagging" field when we store object in bucket.
However, one issue I'm concern is how to rollout this change. Before this change, janitor state objects don't have any tag, which means that when deleting expired buckets, those objects might be deleted together.
From my perspective, it should not be a big issue:
1. Only resources in GlobalTypeList https://github.com/youyangl/boskos/blob/8ed406269010654ac4a50e70e62abf17864a7ef5/aws-janitor/resources/list.go#L107 will be influenced by this issue.
2. Even though the janitor state object got deleted unintentionally, the next time when we run the janitor, the new object will be generated with tag. The "firstseen" map will be regenerated using the timestamp of the new run. I don't think it will cause too much quota issue.

aws-janitor/resources/s3_bucket.go
```diff
@@ -0,0 +1,150 @@
...
+				logger.Warningf("ListObjectsV2Pages failed for bucket: %s : %v", bucketName, err)
+				continue
+			}
+		// If the bucket stores janitor state objects, not to delete it.
+		isStateBucket := false
+		for _, objectKey := range objectsToDelete{
+			if isStateBucket {
+				break
+			}
+			resp, err := svc.GetObjectTagging(&s3.GetObjectTaggingInput{Bucket: aws.String(bucketName), Key: aws.String(*objectKey)})
```
chemikadze, line 81: Feels a little bit too complex and heavy weight, also buckets can be arbitrarily large... Also if it's a new bucket protection probably would work anyway. I guess we can assume state buckets are properly tagged with do not delete tags then and add warning to the flag doc.

aws-janitor/resources/set.go
chemikadze: Taking the feature is behind a feature flag, we probably can just be explicit in flag docs about need to tag state buckets with do not delete tags. Any heuristics that would work in common case feels complex and fragile, and there will be likely edge cases we don't envision anyway that would cause surprising behavior.

cmd/aws-janitor/main.go
```diff
@@ -191,8 +193,19 @@ func markAndSweep(opts resources.Options, region string) error {
 	}
 
 	sweepCount = res.MarkComplete()
-	if err := res.Save(opts.Session, s3p); err != nil {
-		return errors.Wrapf(err, "Error saving %q", *path)
+
+	saveSet := true
+	if *enableS3BucketClean{
+		if !s3path.CheckBucketExist(opts.Session, opts.Region, s3p.Bucket) {
```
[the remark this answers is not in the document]
chemikadze: It may be even better for explicitly fail - user probably will want to know that janitor kills its own state bucket, and terminating without error makes it hard to miss.

chemikadze (changes requested):

aws-janitor/resources/set.go
youyangl: That make sense, we can check the bucket tag every time we set the "path" flag in aws-janitor/main.go, only when the bucket is not managed by janitor, we can do the deletion.

cmd/aws-janitor/main.go
```diff
@@ -162,7 +165,15 @@ func markAndSweep(opts resources.Options, region string) error {
 	if err != nil {
 		return errors.Wrapf(err, "-path %q isn't a valid S3 path", *path)
 	}
-
+	if *enableStateBucketCheck {
```
youyangl, line 168: A state bucket must be tagged with exclude-tags, otherwise, we just return error to avoid deleting it.

cmd/aws-janitor/main.go
```diff
@@ -42,7 +42,7 @@ import (
 var (
 	maxTTL                     = flag.Duration("ttl", 24*time.Hour, "Maximum time before attempting to delete a resource. Set to 0s to nuke all non-default resources.")
 	region                     = flag.String("region", "", "The region to clean (otherwise defaults to all regions)")
-	path                       = flag.String("path", "", "S3 path for mark data (required when -all=false)")
+	path                       = flag.String("path", "", "S3 path for mark data, the bucket must be marked with excludeTags (required when -all=false)")
```
chemikadze, line 45: Feels a bit misleading - it does not have to be marked if bucket clean is off. Add clarification for "if s3 bucket clean is on"?

cmd/aws-janitor/main.go
```diff
@@ -53,6 +53,8 @@ var (
 	enableVPCEndpointsClean    = flag.Bool("enable-vpc-endpoints-clean", false, "If true, clean vpc endpoints.")
 	skipRoute53ManagementCheck = flag.Bool("skip-route53-management-check", false, "If true, skip managed zone check and managed resource name check.")
 	enableDNSZoneClean         = flag.Bool("enable-dns-zone-clean", false, "If true, clean DNS zones.")
+	enableS3BucketClean        = flag.Bool("enable-s3-bucket-clean", false, "If true, clean S3 bucket.")
+	enableStateBucketCheck     = flag.Bool("enable-state-bucket-check", false, "If true, only buckets tagged with exclude-tags can be used to store janitor state data.")
```
chemikadze, line 57: We probably can drop this flag and just always check tagging if s3 bucket clean is requested. Hard to imagine case when s3 clean is on but we don't want to check.

cmd/aws-janitor/main.go
chemikadze: Great idea! Probably can be code comment too.

aws-janitor/resources/s3_bucket.go
```diff
@@ -0,0 +1,141 @@
...
+		return false, err
+	}
+	tags := fromS3Tags(respTags.TagSet)
+	return opts.ManagedPerTags(tags), nil
+}
+
+func (S3Bucket) MarkAndSweep(opts Options, set *Set) error {
+	logger := logrus.WithField("options", opts)
+	if !opts.EnableS3BucketClean {
+		logger.Info("Disable s3 bucket clean")
```
justinsb, line 46: Nit: maybe something like "s3 bucket cleanup is not enabled".  Or is the idea that we shouldn't get here because the set should be empty when `opts.EnableS3BucketClean` is not true?

aws-janitor/resources/s3_bucket.go
```diff
@@ -0,0 +1,141 @@
...
+	resp, err := svc.ListBuckets(&s3.ListBucketsInput{})
+	if err != nil {
+		return err
+	}
+	for _, b := range resp.Buckets {
+		bucketName := *b.Name
+
+		region, err := s3manager.GetBucketRegion(context.Background(), opts.Session, bucketName, opts.Region)
+		if err != nil || region != opts.Region {
+			continue
```
justinsb, line 63: Can we/should we log errors?  Otherwise it's hard to diagnose what's going wrong if buckets are not deleted...

cmd/aws-janitor/main.go
```diff
@@ -53,6 +53,7 @@ var (
 	enableVPCEndpointsClean    = flag.Bool("enable-vpc-endpoints-clean", false, "If true, clean vpc endpoints.")
 	skipRoute53ManagementCheck = flag.Bool("skip-route53-management-check", false, "If true, skip managed zone check and managed resource name check.")
 	enableDNSZoneClean         = flag.Bool("enable-dns-zone-clean", false, "If true, clean DNS zones.")
+	enableS3BucketClean        = flag.Bool("enable-s3-bucket-clean", false, "If true, clean S3 bucket.")
```
justinsb, line 56: nit: buckets

cmd/aws-janitor/main.go
```diff
@@ -162,7 +164,17 @@ func markAndSweep(opts resources.Options, region string) error {
...
-
+	// If we enable S3 bucket clean, the state bucket which stores janitor state data must be tagged with exclude-tags.
+	// Otherwise, if it got cleaned by janitor, the metadata about other resources will also be deleted unintentionally.
+	if *enableS3BucketClean {
+		isManagedBucket, err := resources.IsManagedS3Bucket(opts, region, s3p.Bucket)
+		if err != nil {
+			return errors.Wrapf(err, "Error checking bucket: %s management state", s3p.Bucket)
+		}
+		if isManagedBucket {
+			return errors.Unwrap(fmt.Errorf("State bucket %s, must be tagged with exclude-tags", s3p.Bucket))
```
justinsb, line 175: Nit: why errors.Unwrap?  Can we just `return fmt.Errorf(...)`

cmd/aws-janitor/main.go
```diff
@@ -191,6 +203,7 @@ func markAndSweep(opts resources.Options, region string) error {
 	}
 
 	sweepCount = res.MarkComplete()
+
```
justinsb, line 206: Nit: likely to cause false conflicts in higher-traffic repos.

cmd/aws-janitor-boskos/main.go
```diff
@@ -59,6 +59,7 @@ var (
 	enableVPCEndpointsClean    = flag.Bool("enable-vpc-endpoints-clean", false, "If true, clean vpc endpoints.")
 	skipRoute53ManagementCheck = flag.Bool("skip-route53-management-check", false, "If true, skip managed zone check and managed resource name check.")
 	enableDNSZoneClean         = flag.Bool("enable-dns-zone-clean", false, "If true, clean DNS zones.")
+	enableS3BucketClean        = flag.Bool("enable-s3-bucket-clean", false, "If true, clean S3 bucket.")
```
justinsb, line 62: Nit: buckets

cmd/aws-resources-list/main.go
```diff
@@ -34,6 +34,7 @@ var (
 	enableVPCEndpointsClean    = flag.Bool("enable-vpc-endpoints-clean", false, "If true, clean vpc endpoints.")
 	skipRoute53ManagementCheck = flag.Bool("skip-route53-management-check", false, "If true, skip managed zone check and managed resource name check.")
 	enableDNSZoneClean         = flag.Bool("enable-dns-zone-clean", false, "If true, clean DNS zones.")
+	enableS3BucketClean        = flag.Bool("enable-s3-bucket-clean", false, "If true, clean S3 bucket.")
```
justinsb, line 37: Nit: buckets

aws-janitor/resources/s3_bucket.go
```diff
@@ -0,0 +1,141 @@
...
+	// https://pkg.go.dev/github.com/aws/aws-sdk-go/service/s3#S3.ListBuckets
+	// regardless the region configured in client.
+	resp, err := svc.ListBuckets(&s3.ListBucketsInput{})
+	if err != nil {
+		return err
+	}
+	for _, b := range resp.Buckets {
+		bucketName := *b.Name
+
+		region, err := s3manager.GetBucketRegion(context.Background(), opts.Session, bucketName, opts.Region)
```
justinsb, line 61: Are we missing a call to IsManagedS3Bucket (or am I overlooking it)?

justinsb (commented):
A few nits, but I also don't see the call to IsManagedS3Bucket

aws-janitor/resources/s3_bucket.go
chemikadze: `set.Mark` on line 70 applies same tag checks.

aws-janitor/resources/s3_bucket.go
youyangl: I don't think we should log errors here, the reason is that "ListBuckets" method lists buckets of all regions, and GetBucketRegion can only get regions near "hint region"
https://docs.aws.amazon.com/sdk-for-go/api/service/s3/s3manager/#GetBucketRegion
In this case, if we log error here, most of our log will be filled with the "region error".
To make the error log easier to read, I prefer to let this part fail silently.

aws-janitor/resources/s3_bucket.go
youyangl: We will still get here when the set is empty, 
https://github.com/kubernetes-sigs/boskos/blob/62e7de09c721092f127e01268234897a0b2f3d6b/aws-janitor/resources/clean.go#L58

aws-janitor/resources/s3_bucket.go
justinsb: Ah - thanks!

aws-janitor/resources/s3_bucket.go
justinsb: OK, that makes sense.  A comment as to why might help others avoid adding this in future :-)

justinsb (approved):

PR #155, opened 2022-11-22
Fix a bug for aws iam roles clean up

youyangl:
Many iam roles fail to be deleted due to attached policies they have. Before deleting iam roles, we should also detach those role policies.

### DIFF
--- a/aws-janitor/resources/list.go
+++ b/aws-janitor/resources/list.go
@@ -80,6 +80,7 @@ var RegionalTypeList = []Type{
 	LaunchConfigurations{},
 	LaunchTemplates{},
 	Instances{},
+	VPCEndpoints{},
 	// Addresses
 	NetworkInterfaces{},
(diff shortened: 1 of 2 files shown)


aws-janitor/resources/list.go
```diff
@@ -80,6 +80,7 @@ var RegionalTypeList = []Type{
 	LaunchConfigurations{},
 	LaunchTemplates{},
 	Instances{},
+	VPCEndpoints{},
```
youyangl, line 83: Fix the order issue here, the VPC endpoints should be cleaned ahead of network interfaces and security groups, other wise, some dependency errors might happen.

chemikadze (approved):

justinsb:
LGTM - would have been good to split the VPCEndpoints move into its own commit, but not a blocker

/approve
/lgtm

PR #156, opened 2022-12-07
 Fix a bug for s3 bucket clean

youyangl:


cmd/aws-janitor/main.go
```diff
@@ -167,7 +167,7 @@ func markAndSweep(opts resources.Options, region string) error {
 	// If we enable S3 buckets clean, the state bucket which stores janitor state data must be tagged with exclude-tags.
 	// Otherwise, if it got cleaned by janitor, the metadata about other resources will also be deleted unintentionally.
 	if *enableS3BucketsClean {
-		isManagedBucket, err := resources.IsManagedS3Bucket(opts, region, s3p.Bucket)
+		isManagedBucket, err := resources.IsManagedS3Bucket(opts, s3p.Region, s3p.Bucket)
```
youyangl, line 170: We should use the region parsed from the path flag, otherwise, the janitor state bucket will fail to be found.

justinsb:
Thanks for fixing!

/approve
/lgtm

PR #157, opened 2022-12-19
Reaper: read resource types from Boskos config file

danilo-gemoli:
This PR adds to the `reaper` the feature to read the resource types from a `boskos`'s configuration file.
In this way is easier to keep the two components in sync without having to manually change `reaper`'s deployment every time a new resources is added in `boskos`.

The new `--resouce-type-from-config` switch works together with `--resource-type`: when both of them are specified, the the two lists of resource types are extracted, merged and finally any duplicates is eventually removed.

Furthermore, the file specified by `--resouce-type-from-config` is constantly watched for any new incoming changes so that `reaper` can recompute a new list of resources on the fly.


stevekuznetsov (commented):
The rest of Prow will not allow new but invalid config to crash the service, this is very nice for not accidentally nuking the deployment when a bad config rollout occurs. I would suggest refactoring this to a small `struct` that hides the async re-loading and simply exposes a method `types() []string` that returns the last-loaded config (merged with flags). The `struct` can close over latching on the last successfully-loaded config and log (but not panic) when failing a load.

danilo-gemoli:
@stevekuznetsov nice advice, thanks!

cmd/reaper/types_watcher.go
```diff
@@ -0,0 +1,101 @@
...
+
+				logrus.Infof("new resources: %+v", rtw.mergedTypes)
+			}()
+		})
+		v.WatchConfig()
+	}()
+}
+
+func (rtw *ResourceTypesWatcher) Types() []string {
+	// There is room for a small optimization here:
```
stevekuznetsov, line 76: If your constructor returns an interface (or even just the getter func) you can just have a simple closure that returns the list when no file is given

cmd/reaper/types_watcher.go
```diff
@@ -0,0 +1,101 @@
...
+	return &rt, nil
+}
+
+func (rtw *ResourceTypesWatcher) watchConfig() {
+	go func() {
+		v := viper.New()
+		v.SetConfigFile(rtw.config)
+		v.SetConfigType("yaml")
+		v.OnConfigChange(func(in fsnotify.Event) {
+			// Don't block the handler on reading and parsing the new config
```
stevekuznetsov, line 51: This means that multiple updates in quick succession will race to update the internal state, and you are not guaranteed to converge to the correct state (on disk). Check out the ConfigAgent implementation.

cmd/reaper/types_watcher.go
```diff
@@ -0,0 +1,101 @@
...
+				logrus.Info("a new boskos config is available")
+				newResourceTypes, err := parseResourceTypesFromConfig(rtw.config)
+				if err != nil {
+					logrus.WithError(err).Warn("failed to parse types from a new config")
+					return
+				}
+				newMergedTypes := mergeResourceTypes(newResourceTypes, rtw.providedTypes)
+
+				// Keep the critical region as small as possible
+				func() {
```
stevekuznetsov, line 62: nit: no need for the anonymous function (or block), just

```go
rtw.l.Lock()
rtw.mergedTypes = newMergedTypes
rtw.l.Unlock()
```

cmd/reaper/types_watcher.go
```diff
@@ -0,0 +1,101 @@
...
+		config:        config,
+		providedTypes: providedTypes,
+		l:             &sync.Mutex{},
+	}
+
+	if config != "" {
+		// Read and parse the types for the first time
+		// In this way 'ResourceTypesWatcher.Types()' returns the correct result
+		// even if the config is never updated
+		configResources, err := parseResourceTypesFromConfig(rTypesConfig)
```
stevekuznetsov, line 32: nit: DRY this out :)

cmd/reaper/types_watcher.go
```diff
@@ -0,0 +1,101 @@
...
+	// There is room for a small optimization here:
+	// if the config file was never provided then no watcher is run
+	// and mergeTypes can be safely read without locking on this Mutex
+	rtw.l.Lock()
+	defer rtw.l.Unlock()
+	return rtw.mergedTypes
+}
+
+func mergeResourceTypes(a, b []string) []string {
+	return sets.NewString().Insert(a...).Insert(b...).List()
```
stevekuznetsov, line 85: nit: can be `sets.NewString(a...).Insert(b...)`

cmd/reaper/types_watcher.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/fsnotify/fsnotify"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/viper"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"sigs.k8s.io/boskos/common"
+)
+
+type ResourceTypesWatcher struct {
+	config        string
+	providedTypes []string
```
stevekuznetsov, line 16: nit: if you need this as a `sets.String` on each load, might as well keep that around instead of re-allocating every time

cmd/reaper/types_watcher.go
```diff
@@ -0,0 +1,101 @@
...
+	mergedTypes   []string
+	l             sync.Locker
+}
+
+func NewResourceTypesWatcher(providedTypes []string, config string) (*ResourceTypesWatcher, error) {
+	rt := ResourceTypesWatcher{
+		config:        config,
+		providedTypes: providedTypes,
+		l:             &sync.Mutex{},
```
stevekuznetsov, line 25: nit: use a RWMutex so as to not serialize access

cmd/reaper/resource_types.go
```diff
@@ -0,0 +1,108 @@
...
+	rw          sync.RWMutex
+}
+
+func NewResourceTypes(providedTypes []string, config string) (ResourceTypes, error) {
+	if config == "" {
+		return &resourceTypesStatic{providedTypes: providedTypes}, nil
+	}
+
+	rtw := resourceTypesWatcher{
+		resourceTypesStatic: resourceTypesStatic{
```
stevekuznetsov, line 37: why not hold on to `sets.NewString(providedTypes)` instead of these layers of indirection?

stevekuznetsov (approved):

cmd/reaper/resource_types.go
danilo-gemoli: I'm using a `sets.String` as a temporary, transient area of storage because it's a convenient mean to merge the `types` from both the config file and the command line.
I won't store the set as a member of `ResourceTypes` struct cause it would be difficult to recompute the `types` when the config file changes.
Let's picture the following use case:
- `providedTypes` is `{"a", "b"}`
- `types` from config are `{"b", "d"}`
- the resulting `types` would be `{"a", "b", "d"}`
- now the config changes into `{"b", "e"}`
- at this stage would be not so obvious to get the correct result `{"a", "b", "e"}` starting only from the last known set `{"a", "b", "d"}`. It would be much easier just to `sets.NewString(providedTypes...).Insert(newConfigTypes...).List()`

I'm obviously open to changes if you see something that I don't.

danilo-gemoli:
@stevekuznetsov @fejta is there anything else I can do to have this merged?

cmd/reaper/resource_types.go
stevekuznetsov: Store the set of `providedTypes`, not the merged set.

danilo-gemoli: Is this (or the like) what you mean?
```golang
func (rtw *resourceTypesWatcher) Types() []string {
	types, err := rtw.mergeTypesFromConfig()
	if err != nil {
		logrus.Warnf("merge types: %s", err.Error())
		return rtw.providedTypes
	}
	return types
}
```
That is: reading from the file and therefore merging with the provided ones without bothering at all with the "watcher stuff"?

stevekuznetsov: No - I am saying just that the whole `resourceTypesStatic` type is superfluous - you could just as well write

```go
type resourceTypesWatcher struct {
	providedTypes sets.String
	config      string
	mergedTypes []string
	rw          sync.RWMutex
}
```

danilo-gemoli: Removed, it looks like yours.

stevekuznetsov (approved):

TaichiHo:
Should we have the same feature for janitor? Since they often go hand in hand?

danilo-gemoli:
if @stevekuznetsov agrees, why not

PR #158, opened 2023-01-05
Disassociate elastic IP address when detaching Internet Gateway

youyangl:
According to AWS tutorial:
https://docs.aws.amazon.com/vpc/latest/userguide/VPC_Internet_Gateway.html#detach-igw The elastic ip should be disassociated before detaching vpc internet gateway.

### DIFF
--- a/aws-janitor/resources/internet_gateways.go
+++ b/aws-janitor/resources/internet_gateways.go
@@ -77,6 +77,37 @@ func (InternetGateways) MarkAndSweep(opts Options, set *Set) error {
 				break
 			}
 
+			var publicIPsToRelease []*string
+
+			pageFunc := func(page *ec2.DescribeNetworkInterfacesOutput, _ bool) bool {
+				for _, eni := range page.NetworkInterfaces {
+					publicIPsToRelease = append(publicIPsToRelease, eni.Association.PublicIp)
+				}
+				return true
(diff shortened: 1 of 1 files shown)


elton-d (approved):

justinsb:
Thanks for fixing!

/approve
/lgtm

justinsb:
Thanks!

/approve
/lgtm

aws-janitor/resources/internet_gateways.go
```diff
@@ -77,6 +77,37 @@ func (InternetGateways) MarkAndSweep(opts Options, set *Set) error {
 				break
 			}
 
+			var publicIPsToRelease []*string
+
+			pageFunc := func(page *ec2.DescribeNetworkInterfacesOutput, _ bool) bool {
+				for _, eni := range page.NetworkInterfaces {
+					publicIPsToRelease = append(publicIPsToRelease, eni.Association.PublicIp)
```
chemikadze, line 84: I believe association can be nil, causing cleanup to panic in certain cases. This also made difficult by code in main() that appears to hide panic logging as it calls explicit os.Exit.

Also, deleting network interfaces should disassociate automatically, so likely we don't need this change - I'm suspecting that ENIs are held by untagged ELBs.

PR #159, opened 2023-01-12
Skip some resource record set types when deleting resource records

youyangl:
Currently, we have error log:
unable to delete DNS records: InvalidChangeBatch: [A HostedZone must contain at least one NS record for the zone itself., A HostedZone must contain exactly one SOA record.]

Which means that the type "NS" and "SOA" should not be deleted.

### DIFF
--- a/aws-janitor/resources/route53.go
+++ b/aws-janitor/resources/route53.go
@@ -94,9 +94,11 @@ func route53ResourceRecordSetsForZone(opts Options, logger logrus.FieldLogger, s
 			if !set.Mark(opts, o, nil, zoneTags) {
 				continue
 			}
-			logger.Warningf("%s: deleting %T: %s", o.ARN(), rrs, *rrs.Name)
-			if !opts.DryRun {
-				toDelete = append(toDelete, o)
+			if _, ok := opts.SkipResourceRecordSetTypes[*rrs.Type]; !ok {
+				logger.Warningf("%s: deleting %T: %s", o.ARN(), rrs, *rrs.Name)
+				if !opts.DryRun {
+					toDelete = append(toDelete, o)
+				}
 			}
 		}
 		return true
@@ -177,7 +179,7 @@ func (rrs Route53ResourceRecordSets) MarkAndSweep(opts Options, set *Set) error 
 			}
 
 			if _, err := svc.ChangeResourceRecordSets(deleteReq); err != nil {
-				logger.Warningf("unable to delete DNS records: %v", err)
+				logger.Warningf("unable to delete DNS records for zone %v with error: %v", aws.StringValue(zone.Id), err)
 			}
 		}
 
@@ -186,7 +188,7 @@ func (rrs Route53ResourceRecordSets) MarkAndSweep(opts Options, set *Set) error 
 				Id: zone.Id,
 			}
 			if _, err := svc.DeleteHostedZone(deleteHostZoneReq); err != nil {
-				logger.Warningf("unable to delete DNS zone: %v", err)
+				logger.Warningf("unable to delete DNS zone %v with error: %v", aws.StringValue(zone.Id), err)
 			}
 		}
 	}
--- a/cmd/aws-janitor-boskos/main.go
+++ b/cmd/aws-janitor-boskos/main.go
@@ -61,10 +61,12 @@ var (
 	enableDNSZoneClean         = flag.Bool("enable-dns-zone-clean", false, "If true, clean DNS zones.")
 	enableS3BucketsClean       = flag.Bool("enable-s3-buckets-clean", false, "If true, clean S3 buckets.")
 
-	excludeTags common.CommaSeparatedStrings
-	includeTags common.CommaSeparatedStrings
-	excludeTM   resources.TagMatcher
-	includeTM   resources.TagMatcher
+	excludeTags                common.CommaSeparatedStrings
+	includeTags                common.CommaSeparatedStrings
+	skipResourceRecordSetTypes common.CommaSeparatedStrings
+
+	excludeTM resources.TagMatcher
+	includeTM resources.TagMatcher
 
 	instrumentationOptions prowflagutil.InstrumentationOptions
 
@@ -90,6 +92,7 @@ func init() {
 		"Resources with any of these tags will not be managed by the janitor. Given as a comma-separated list of tags in key[=value] format; excluding the value will match any tag with that key. Keys can be repeated.")
 	flag.Var(&includeTags, "include-tags",
 		"Resources must include all of these tags in order to be managed by the janitor. Given as a comma-separated list of tags in key[=value] format; excluding the value will match any tag with that key. Keys can be repeated.")
+	flag.Var(&skipResourceRecordSetTypes, "skip-resource-record-set-types", "A list of resource record types which should not be deleted, splitted using comma.")
 
 	prometheus.MustRegister(cleaningTimeHistogram)
 	prometheus.MustRegister(sweepsGauge)
@@ -185,6 +188,19 @@ func cleanResource(res *common.Resource) error {
 	if err != nil {
 		return errors.Wrap(err, "Failed retrieving account")
 	}
+	skipResourceRecordSetTypesSet := map[string]bool{}
+	// A HostedZone must contain at least one NS record for the zone itself.
+	// A HostedZone must contain exactly one SOA record.
+	// Thus, by default, we should not delete record set with type 'NS' or 'SOA'.
+	if len(skipResourceRecordSetTypes) == 0 {
+		err := skipResourceRecordSetTypes.Set("SOA,NS")
+		if err != nil {
+			logrus.Fatalf("Error setting --skip-resource-record-set-types: %v", err)
+		}
+	}
+	for _, resourceRecordType := range skipResourceRecordSetTypes {
+		skipResourceRecordSetTypesSet[resourceRecordType] = true
+	}
 	opts := resources.Options{
 		Session:                    s,
 		Account:                    acct,
@@ -198,6 +214,7 @@ func cleanResource(res *common.Resource) error {
 		SkipRoute53ManagementCheck: *skipRoute53ManagementCheck,
 		EnableDNSZoneClean:         *enableDNSZoneClean,
 		EnableS3BucketsClean:       *enableS3BucketsClean,
+		SkipResourceRecordSetTypes: skipResourceRecordSetTypesSet,
 	}
 
 	logrus.WithField("name", res.Name).Info("beginning cleaning")
--- a/cmd/aws-janitor/main.go
+++ b/cmd/aws-janitor/main.go
@@ -55,8 +55,9 @@ var (
 	enableDNSZoneClean         = flag.Bool("enable-dns-zone-clean", false, "If true, clean DNS zones.")
 	enableS3BucketsClean       = flag.Bool("enable-s3-buckets-clean", false, "If true, clean S3 buckets.")
 
-	excludeTags common.CommaSeparatedStrings
-	includeTags common.CommaSeparatedStrings
+	excludeTags                common.CommaSeparatedStrings
+	includeTags                common.CommaSeparatedStrings
+	skipResourceRecordSetTypes common.CommaSeparatedStrings
 
 	sweepCount int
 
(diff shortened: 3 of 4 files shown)


aws-janitor/resources/route53.go
```diff
@@ -96,7 +96,9 @@ func route53ResourceRecordSetsForZone(opts Options, logger logrus.FieldLogger, s
 			}
 			logger.Warningf("%s: deleting %T: %s", o.ARN(), rrs, *rrs.Name)
 			if !opts.DryRun {
-				toDelete = append(toDelete, o)
+				if _, ok := opts.SkipResourceRecordSetTypes[*rrs.Type]; !ok{
+					toDelete = append(toDelete, o)
```
youyangl, line 100: Another thing we should keep in mind is that according to AWS tutorial:
https://docs.aws.amazon.com/cli/latest/reference/route53/change-resource-record-sets.html
Warning
To delete the resource record set that is associated with a traffic policy instance, use [DeleteTrafficPolicyInstance](https://docs.aws.amazon.com/Route53/latest/APIReference/API_DeleteTrafficPolicyInstance.html) . Amazon Route 53 will delete the resource record set automatically. If you delete the resource record set by using ChangeResourceRecordSets , Route 53 doesn't automatically delete the traffic policy instance, and you'll continue to be charged for it even though it's no longer in use.

I will add corresponding logic in follow up changes

cmd/aws-janitor-boskos/main.go
```diff
@@ -60,6 +61,7 @@ var (
 	skipRoute53ManagementCheck = flag.Bool("skip-route53-management-check", false, "If true, skip managed zone check and managed resource name check.")
 	enableDNSZoneClean         = flag.Bool("enable-dns-zone-clean", false, "If true, clean DNS zones.")
 	enableS3BucketsClean       = flag.Bool("enable-s3-buckets-clean", false, "If true, clean S3 buckets.")
+	skipResourceRecordSetTypes = flag.String("skip-resource-record-set-types", "NS,SOA", "A list of resource record types which should not be deleted, splitted using comma.")
```
youyangl, line 64: We set the default value to be "NS,SOA" here, maybe we should leave it empty since it influences all users?

chemikadze, line 64: Would common.CommaSeparatedStrings fit this better?

cmd/aws-janitor/main.go
```diff
@@ -54,6 +55,7 @@ var (
 	skipRoute53ManagementCheck = flag.Bool("skip-route53-management-check", false, "If true, skip managed zone check and managed resource name check.")
 	enableDNSZoneClean         = flag.Bool("enable-dns-zone-clean", false, "If true, clean DNS zones.")
 	enableS3BucketsClean       = flag.Bool("enable-s3-buckets-clean", false, "If true, clean S3 buckets.")
+	skipResourceRecordSetTypes = flag.String("skip-resource-record-set-types", "NS,SOA", "A list of resource record types which should not be deleted, splitted using comma.")
```
chemikadze, line 58: same

aws-janitor/resources/route53.go
```diff
@@ -96,7 +96,9 @@ func route53ResourceRecordSetsForZone(opts Options, logger logrus.FieldLogger, s
...
+				}
 			}
 		}
 		return true
```
chemikadze, line 104: (can't comment on line 182, 191)

"unable to delete" log lines seem to lack context about what record / zone caused the issue - would it make sense to add resource name / arn to the error message?

aws-janitor/resources/route53.go
```diff
@@ -96,7 +96,9 @@ func route53ResourceRecordSetsForZone(opts Options, logger logrus.FieldLogger, s
 			}
 			logger.Warningf("%s: deleting %T: %s", o.ARN(), rrs, *rrs.Name)
 			if !opts.DryRun {
-				toDelete = append(toDelete, o)
+				if _, ok := opts.SkipResourceRecordSetTypes[*rrs.Type]; !ok{
```
chemikadze, line 99: I have a feeling that we are practically trying to un-hardcode `resourceRecordSetIsManaged` by this change, is there some better way of telling which records are managed / "non-required" than just comparing their type?

Also, if reason why we adding this change is to better select managed records, should condition go inside `resourceRecordSetIsManaged`? Does `SkipRoute53ManagementCheck` make sense in context of what we've learned with NS and SOA records?

chemikadze: > Does SkipRoute53ManagementCheck make sense in context of what we've learned with NS and SOA records?

Taking this back, looks like this is also suppressing regex matching which makes sense in either case. Though "management check" for records seem to mix "records that are managed because they look like something created by tests" and "required records that every zone should have". Should effect of SkipRoute53ManagementCheck apply only to former, but not latter? Should we have check for the latter active regardless of SkipRoute53ManagementCheck?

youyangl: 1. function: zoneIsManaged only accepts dns zones with given name formats. For our usage, we need to skip that check since many of our dns zones that need to be handled by janitor don't have such kind of name formats.

2. SkipResourceRecordSetTypes filters out resource record sets with type: "NS" and "SOA", the reason why we need to so is that at line 182, when trying to delete resource record sets, the deletion fail with error log:
"unable to delete DNS records: InvalidChangeBatch: [A HostedZone must contain at least one NS record for the zone itself., A HostedZone must contain exactly one SOA record.]" 
which means that NS and SOA record sets should not be deleted

aws-janitor/resources/route53.go
chemikadze: We can skip it for now if traffic policies are not being created.

aws-janitor/resources/route53.go
```diff
@@ -177,7 +179,7 @@ func (rrs Route53ResourceRecordSets) MarkAndSweep(opts Options, set *Set) error
 			}
 
 			if _, err := svc.ChangeResourceRecordSets(deleteReq); err != nil {
-				logger.Warningf("unable to delete DNS records: %v", err)
+				logger.Warningf("unable to delete DNS records for zone: %v with error: %v", zone.Id, err)
```
chemikadze, line 182: nit: remove colon in `zone:`, as it's all part of one message?

aws-janitor/resources/route53.go
```diff
@@ -186,7 +188,7 @@ func (rrs Route53ResourceRecordSets) MarkAndSweep(opts Options, set *Set) error
 				Id: zone.Id,
 			}
 			if _, err := svc.DeleteHostedZone(deleteHostZoneReq); err != nil {
-				logger.Warningf("unable to delete DNS zone: %v", err)
+				logger.Warningf("unable to delete DNS zone: %v with error: %v", zone.Id, err)
```
chemikadze, line 191: same

cmd/aws-janitor-boskos/main.go
```diff
@@ -90,6 +92,7 @@ func init() {
 		"Resources with any of these tags will not be managed by the janitor. Given as a comma-separated list of tags in key[=value] format; excluding the value will match any tag with that key. Keys can be repeated.")
 	flag.Var(&includeTags, "include-tags",
 		"Resources must include all of these tags in order to be managed by the janitor. Given as a comma-separated list of tags in key[=value] format; excluding the value will match any tag with that key. Keys can be repeated.")
+	flag.Var(&skipResourceRecordSetTypes, "skip-resource-record-set-types", "A list of resource record types which should not be deleted, splitted using comma.")
```
chemikadze, line 95: I think we have to initialize default value for the flag / var to `NS, SOA`, otherwise `--skip-route53-management-check` would not work by default without explicit `--skip-resource-record-set-types`.

cmd/aws-janitor/main.go
```diff
@@ -77,6 +78,7 @@ func init() {
 		"Resources with any of these tags will not be managed by the janitor. Given as a comma-separated list of tags in key[=value] format; excluding the value will match any tag with that key. Keys can be repeated.")
 	flag.Var(&includeTags, "include-tags",
 		"Resources must include all of these tags in order to be managed by the janitor. Given as a comma-separated list of tags in key[=value] format; excluding the value will match any tag with that key. Keys can be repeated.")
+	flag.Var(&skipResourceRecordSetTypes, "skip-resource-record-set-types", "A list of resource record types which should not be deleted, splitted using comma.")
```
chemikadze, line 81: Same.

aws-janitor/resources/route53.go
chemikadze: Semantics we are getting just feels a bit confusing, "managed check" for records combines both type and name validation and type validation also happens to have important side effect of making deletions actually valid, and adding separate "skip resource record set types" adds one more place where type verification is happening.

I would imagine the following flags as potentially more flexible arrangement instead of complete managed check disabling and additional type check:

 - string list for managed zone regexes (defaulted to test-cncf-aws.k8s.io.)
 - string list for managed record regexes (defaulted to contents of managedNameRegexes)
 - string list for record types considered for deletion (defaulted to A)

Though if we already went ahead with disabling managed check altogether and no concerns were raised in previous reviews, I'm probably overthinking it and we can keep it as is.

cmd/aws-janitor-boskos/main.go
chemikadze: `--skip-route53-management-check=true` would not work without it, so setting by default is preferred, I think. It would not do anything with default `--skip-route53-management-check=false` anyway.

chemikadze (approved):

cmd/aws-janitor-boskos/main.go
youyangl: We can't set default value at this line, the reason is that flag.Var only supports:
flag.Var(&flagVal, "name", "help message for flagname").
Nor can we set default value when initializing the var: skipResourceRecordSetTypes, the reason is that for 
common.CommaSeparatedStrings, if it initialized with non-empty value, the "Set" method will return error once being called.
https://github.com/kubernetes-sigs/boskos/blob/3e30fa430bc9bd77ea67d3a70bc74ce43b00190c/common/common.go#L211

What we can do is that before converting the string array to set, we can set its value if its empty.
See line: 193

cmd/aws-janitor/main.go
youyangl: Same above

aws-janitor/resources/route53.go
```diff
@@ -96,7 +96,9 @@ func route53ResourceRecordSetsForZone(opts Options, logger logrus.FieldLogger, s
 			}
 			logger.Warningf("%s: deleting %T: %s", o.ARN(), rrs, *rrs.Name)
```
justinsb, line 97: We should move the Skip check so it doesn't log this if it's not actually going to delete the record.

aws-janitor/resources/route53.go
```diff
@@ -177,7 +179,7 @@ func (rrs Route53ResourceRecordSets) MarkAndSweep(opts Options, set *Set) error
 			}
 
 			if _, err := svc.ChangeResourceRecordSets(deleteReq); err != nil {
-				logger.Warningf("unable to delete DNS records: %v", err)
+				logger.Warningf("unable to delete DNS records for zone %v with error: %v", zone.Id, err)
```
justinsb, line 182: I think this is going to log the pointer, you probably want `aws.StringValue(zone.Id)`

aws-janitor/resources/route53.go
```diff
@@ -186,7 +188,7 @@ func (rrs Route53ResourceRecordSets) MarkAndSweep(opts Options, set *Set) error
 				Id: zone.Id,
 			}
 			if _, err := svc.DeleteHostedZone(deleteHostZoneReq); err != nil {
-				logger.Warningf("unable to delete DNS zone: %v", err)
+				logger.Warningf("unable to delete DNS zone %v with error: %v", zone.Id, err)
```
justinsb, line 191: Likewise here, I think this will log the pointer

cmd/aws-janitor/main.go
```diff
@@ -130,7 +132,13 @@ func main() {
 		logrus.Errorf("Error parsing --include-tags: %v", err)
 		runtime.Goexit()
 	}
-
+	skipResourceRecordSetTypesSet := map[string]bool{}
+	if len(skipResourceRecordSetTypes) == 0 {
+		skipResourceRecordSetTypes.Set("SOA,NS")
```
justinsb, line 137: A comment that these cannot be deleted would probably be great, to justify the default.

justinsb:
Thanks for tweaks, lgtm

/approve
/lgtm

justinsb:
Thanks!

/approve
/lgtm
/ok-to-test

PR #160, opened 2023-01-19
Add cleanup for VPC load balancers

Amulyam24:
Add cleanup of load balancers for VPC resources

### DIFF
--- a/internal/ibmcloud-janitor/resources/types.go
+++ b/internal/ibmcloud-janitor/resources/types.go
@@ -45,6 +45,7 @@ var PowervsResources = []Resource{
 
 var VpcResources = []Resource{
 	VPCInstance{},
+	VPCLoadBalancer{},
 	VPCNetwork{},
 	VPCs{},
 }
@@ -74,6 +75,9 @@ type VPC interface {
 	GetSubnetPublicGateway(options *vpcv1.GetSubnetPublicGatewayOptions) (*vpcv1.PublicGateway, *core.DetailedResponse, error)
 	DeletePublicGateway(options *vpcv1.DeletePublicGatewayOptions) (*core.DetailedResponse, error)
 	UnsetSubnetPublicGateway(options *vpcv1.UnsetSubnetPublicGatewayOptions) (*core.DetailedResponse, error)
+	DeleteLoadBalancer(options *vpcv1.DeleteLoadBalancerOptions) (*core.DetailedResponse, error)
+	ListLoadBalancers(options *vpcv1.ListLoadBalancersOptions) (*vpcv1.LoadBalancerCollection, *core.DetailedResponse, error)
+	GetLoadBalancer(options *vpcv1.GetLoadBalancerOptions) (result *vpcv1.LoadBalancer, response *core.DetailedResponse, err error)
 }
 
 type ServiceIDClient interface {
--- a/internal/ibmcloud-janitor/resources/vpc_lbs.go
+++ b/internal/ibmcloud-janitor/resources/vpc_lbs.go
@@ -1,0 +1,119 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package resources
+
+import (
+	"strings"
+	"time"
+
+	"github.com/IBM/vpc-go-sdk/vpcv1"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+type VPCLoadBalancer struct{}
+
+var (
+	lbDeletionTimeout = time.Minute * 4
+	lbPollingInterval = time.Second * 30
+	resourceLogger    *logrus.Entry
+)
+
+// Cleans up the load balancers in a given region
+func (VPCLoadBalancer) cleanup(options *CleanupOptions) error {
+	resourceLogger = logrus.WithFields(logrus.Fields{"resource": options.Resource.Name})
+	resourceLogger.Info("Cleaning up the load balancers")
+	client, err := NewVPCClient(options)
+	if err != nil {
+		return errors.Wrap(err, "couldn't create VPC client")
+	}
+
+	var deletedLBList []string
+	var errs []error
+
+	f := func(start string) (bool, string, error) {
+		listLbOpts := &vpcv1.ListLoadBalancersOptions{}
+		if start != "" {
+			listLbOpts.Start = &start
+		}
+
+		loadBalancers, _, err := client.ListLoadBalancers(listLbOpts)
+		if err != nil {
+			return false, "", errors.Wrap(err, "failed to list the load balancers")
+		}
+
+		if loadBalancers == nil || len(loadBalancers.LoadBalancers) <= 0 {
+			resourceLogger.Info("there are no available load balancers to delete")
+			return true, "", nil
+		}
+
+		for _, lb := range loadBalancers.LoadBalancers {
+			if *lb.ResourceGroup.ID == client.ResourceGroupID {
+				if _, err := client.DeleteLoadBalancer(&vpcv1.DeleteLoadBalancerOptions{
+					ID: lb.ID,
+				}); err != nil {
+					resourceLogger.WithField("name", *lb.Name).Error("failed to delete load balancer")
+					errs = append(errs, err)
+					continue
+				}
+				deletedLBList = append(deletedLBList, *lb.ID)
+				resourceLogger.WithField("name", *lb.Name).Info("load balancer deletetion triggered")
+			}
+		}
+
+		if loadBalancers.Next != nil && *loadBalancers.Next.Href != "" {
+			return false, *loadBalancers.Next.Href, nil
+		}
+
+		return true, "", nil
+	}
+
+	if err = pagingHelper(f); err != nil {
+		errs = append(errs, errors.Wrapf(err, "failed to delete the load balancers"))
+	}
+
+	// check if the LBs are properly deleted
+	checkLBs(deletedLBList, client, &errs)
+
+	if len(errs) > 0 {
+		return kerrors.NewAggregate(errs)
+	}
+
+	resourceLogger.Info("Successfully deleted the load balancers")
+	return nil
+}
+
+func checkLBs(list []string, client *IBMVPCClient, errs *[]error) {
+	check := func(id string) error {
+		return wait.PollImmediate(lbPollingInterval, lbDeletionTimeout, func() (bool, error) {
+			_, _, err := client.GetLoadBalancer(&vpcv1.GetLoadBalancerOptions{ID: &id})
+			if err != nil && strings.Contains(err.Error(), "cannot be found") {
+				return true, nil
+			}
+			return false, err
+		})
+	}
+
+	for _, lb := range list {
+		if err := check(lb); err != nil {
+			resourceLogger.WithField("ID", lb).Error("failed to check the deletion of load balancer")
+			*errs = append(*errs, err)
+		}
+	}
+}
--- a/internal/ibmcloud-janitor/resources/util.go
+++ b/internal/ibmcloud-janitor/resources/util.go
@@ -1,0 +1,67 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package resources
+
+import (
+	"net/url"
+
+	"github.com/pkg/errors"
+)
+
+// pagingHelper is used while listing resources. It parses and fetches the start token for
+// getting the next set of resources if nextURL is returned by func f.
+// func f has start as a parameter and returns the following:
+// isDone bool - true denotes that iterating is not needed as there is no next set of resources
+// nextURL string - denotes a URL for a page of resources which is parsed for fetching start token for next iteration
+// e error - will break and return error if e is not nil
+func pagingHelper(f func(string) (bool, string, error)) (err error) {
+	start := ""
+
+	getStartToken := func(nextURL string) (string, error) {
+		url, err := url.Parse(nextURL)
+		if err != nil || url == nil {
+			return "", errors.Wrapf(err, "failed to parse next url for getting next resources")
+		}
+
+		start := url.Query().Get("start")
+		return start, nil
+	}
+
+	for {
+		isDone, nextURL, e := f(start)
+
+		if e != nil {
+			err = e
+			break
+		}
+
+		if isDone {
+			break
+		}
+
+		if nextURL != "" {
+			start, err = getStartToken(nextURL)
+			if err != nil {
+				break
+			}
+		} else {
+			break
+		}
+	}
+
+	return
+}
--- a/internal/ibmcloud-janitor/resources/vpc_client.go
+++ b/internal/ibmcloud-janitor/resources/vpc_client.go
@@ -78,6 +78,18 @@ func (c *IBMVPCClient) UnsetSubnetPublicGateway(options *vpcv1.UnsetSubnetPublic
 	return c.vpcService.UnsetSubnetPublicGateway(options)
 }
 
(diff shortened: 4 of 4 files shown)


internal/ibmcloud-janitor/resources/types.go
```diff
@@ -47,6 +47,7 @@ var VpcResources = []Resource{
 	VPCInstance{},
 	VPCNetwork{},
 	VPCs{},
+	VPCLoadBalancer{},
```
mkumatag, line 50: check the order, I believe loadbalancers will have subnets and need to be deleted before cleaning up the networks

need your help to review this PR further @dharaneeshvrd 
/cc @Prajyot-Parab

internal/ibmcloud-janitor/resources/vpc_lbs.go
```diff
@@ -0,0 +1,51 @@
...
+// Cleans up the load balancers in a given region
+func (VPCLoadBalancer) cleanup(options *CleanupOptions) error {
+	resourceLogger := logrus.WithFields(logrus.Fields{"resource": options.Resource.Name})
+	resourceLogger.Info("Cleaning up the load balancers")
+	client, err := NewVPCClient(options)
+	if err != nil {
+		return errors.Wrap(err, "couldn't create VPC client")
+	}
+
+	loadBalancers, _, err := client.ListLoadBalancers(&vpcv1.ListLoadBalancersOptions{})
```
mkumatag, line 36: see if this accepts the resource group or any other options to list whatever needs to be deleted instead of listing entire thing.

go.mod
```diff
@@ -3,10 +3,10 @@ module sigs.k8s.io/boskos
...
 
 require (
-	github.com/IBM-Cloud/power-go-client v1.2.1
-	github.com/IBM/go-sdk-core/v5 v5.10.2
-	github.com/IBM/platform-services-go-sdk v0.28.6
-	github.com/IBM/vpc-go-sdk v0.27.0
+	github.com/IBM-Cloud/power-go-client v1.2.2
+	github.com/IBM/go-sdk-core/v5 v5.12.0
+	github.com/IBM/platform-services-go-sdk v0.31.3
+	github.com/IBM/vpc-go-sdk v0.30.0
```
mkumatag, line 9: this update can go in a different PR to avoid dependency on the other maintainer for approving the code.

mkumatag (changes requested):

internal/ibmcloud-janitor/resources/vpc_lbs.go
Amulyam24: For options, we only have these
```
type ListLoadBalancersOptions struct {
    // A server-provided token determining what resource to start the page on.
    Start *string `json:"start,omitempty"`

    // The number of resources to return on a page.
    Limit *int64 `json:"limit,omitempty"`

    // Allows users to set headers on API requests
    Headers map[string]string
}
```

Amulyam24: Initial plan was to check with resource group similar to deleting instances & networks, but it isn't available

dharaneeshvrd: Yes listing load balancer is not providing any filters.

If the intention is to delete all the load balancers, then need to take care of paging as well.
https://github.com/openshift/hypershift/blob/main/cmd/infra/powervs/destroy.go#L735

internal/ibmcloud-janitor/resources/types.go
dharaneeshvrd: Under subnets we will have load balancer as an attached resource, need to cleanup load balancer, before proceeding the deletion of subnet and then vpc.

Amulyam24: Changed it.

mkumatag (changes requested):

internal/ibmcloud-janitor/resources/vpc_lbs.go
Amulyam24: Updated with changes, PTAL.

internal/ibmcloud-janitor/resources/vpc_lbs.go
```diff
@@ -0,0 +1,71 @@
...
+			return false, "", errors.Wrap(err, "failed to list the load balancers")
+		}
+
+		if loadBalancers == nil || len(loadBalancers.LoadBalancers) <= 0 {
+			resourceLogger.Info("there are no available load balancers to delete")
+			return true, "", nil
+		}
+
+		for _, lb := range loadBalancers.LoadBalancers {
+			if lb.ResourceGroup.ID == &client.ResourceGroupID {
```
dharaneeshvrd, line 53: I think need to compare dereferenced value, not the pointer.

Amulyam24: That's right, fixed it.

dharaneeshvrd:
In case of subnet deletion is attempted immediately after this cleanup, need to ensure, load balancer is deleted fully, before proceeding to the subnet or else subnet deletion will fail.
Like this block https://github.com/openshift/hypershift/blob/main/cmd/infra/powervs/destroy.go#L689

mkumatag (changes requested):

internal/ibmcloud-janitor/resources/vpc_lbs.go
```diff
@@ -0,0 +1,85 @@
...
+
+		for _, lb := range loadBalancers.LoadBalancers {
+			if *lb.ResourceGroup.ID == client.ResourceGroupID {
+				if _, err := client.DeleteLoadBalancer(&vpcv1.DeleteLoadBalancerOptions{
+					ID: lb.ID,
+				}); err != nil {
+					return false, "", errors.Wrapf(err, "failed to delete the load balancer %q", *lb.Name)
+				}
+			}
+			checkLb(*lb.ID)
```
dharaneeshvrd, line 73: IMO we can split the deletion and checkLb separately, if we do like this sequentially it is going to take lot of time if we need to delete more LBs.
 
You can trigger the deletion on all the load balancers and process their deletion separately by storing deleted Lb IDs in a list or you can process this block in a goroutine maybe, so it would make things faster. I would prefer the 1st approach.

Prajyot-Parab:
lgtm @dharaneeshvrd can you PTAL once.

internal/ibmcloud-janitor/resources/vpc_lbs.go
```diff
@@ -0,0 +1,116 @@
...
+
+		for _, lb := range loadBalancers.LoadBalancers {
+			if *lb.ResourceGroup.ID == client.ResourceGroupID {
+				if _, err := client.DeleteLoadBalancer(&vpcv1.DeleteLoadBalancerOptions{
+					ID: lb.ID,
+				}); err != nil {
+					return false, "", errors.Wrapf(err, "failed to delete the load balancer %q", *lb.Name)
+				}
+			}
+			lbList[*lb.ID] = false
```
dharaneeshvrd, line 69: I think this line should be inside if.

internal/ibmcloud-janitor/resources/vpc_lbs.go
```diff
@@ -0,0 +1,116 @@
...
+// Cleans up the load balancers in a given region
+func (VPCLoadBalancer) cleanup(options *CleanupOptions) error {
+	resourceLogger := logrus.WithFields(logrus.Fields{"resource": options.Resource.Name})
+	resourceLogger.Info("Cleaning up the load balancers")
+	client, err := NewVPCClient(options)
+	if err != nil {
+		return errors.Wrap(err, "couldn't create VPC client")
+	}
+
+	var lbList map[string]bool
```
dharaneeshvrd, line 43: can we rename it to denote that it contains list of deleted LBs?

internal/ibmcloud-janitor/resources/vpc_lbs.go
```diff
@@ -0,0 +1,116 @@
...
+					list[id] = true
+					continue
+				} else if *lb.ProvisioningStatus == "delete_pending" {
+					break
+				} else {
+					return err
+				}
+			}
+		}
+		fmt.Println("load balancer deletion is pending, checking again in a minute")
```
dharaneeshvrd, line 110: Is it ok to use fmt.Println instead of resourceLogger.Info here?

internal/ibmcloud-janitor/resources/vpc_lbs.go
```diff
@@ -0,0 +1,116 @@
...
+
+	resourceLogger.Info("Successfully deleted the load balancers")
+	return nil
+}
+
+func checkLBs(list map[string]bool, client *IBMVPCClient) error {
+	timeout := 0 * time.Minute
+	for timeout < lbTimeout {
+		for id, lbDeleted := range list {
+			lbID := id
```
dharaneeshvrd, line 97: Is this required?

internal/ibmcloud-janitor/resources/vpc_lbs.go
dharaneeshvrd: It's not initialised, need to use make.

internal/ibmcloud-janitor/resources/vpc_lbs.go
```diff
@@ -0,0 +1,116 @@
...
+		for _, lb := range loadBalancers.LoadBalancers {
+			if *lb.ResourceGroup.ID == client.ResourceGroupID {
+				if _, err := client.DeleteLoadBalancer(&vpcv1.DeleteLoadBalancerOptions{
+					ID: lb.ID,
+				}); err != nil {
+					return false, "", errors.Wrapf(err, "failed to delete the load balancer %q", *lb.Name)
+				}
+			}
+			lbList[*lb.ID] = false
+			resourceLogger.WithField("name", *lb.Name).Info("load balancer deleted")
```
dharaneeshvrd, line 70: can we log as "deletion triggered"? As at this point of time it won't be deleted fully.

Amulyam24: Changed it.

internal/ibmcloud-janitor/resources/vpc_lbs.go
Amulyam24: Changed it.

internal/ibmcloud-janitor/resources/vpc_lbs.go
Amulyam24: yes, fixed it.

internal/ibmcloud-janitor/resources/vpc_lbs.go
Amulyam24: I added this because when I directly assign the id as `GetLoadBalancer(&vpcv1.GetLoadBalancerOptions{ID: &id}`, it was failing with a lint error.

Amulyam24:
Review comments addressed, PTAL.

internal/ibmcloud-janitor/resources/vpc_lbs.go
```diff
@@ -0,0 +1,118 @@
...
+		for _, lb := range loadBalancers.LoadBalancers {
+			if *lb.ResourceGroup.ID == client.ResourceGroupID {
+				if _, err := client.DeleteLoadBalancer(&vpcv1.DeleteLoadBalancerOptions{
+					ID: lb.ID,
+				}); err != nil {
+					return false, "", errors.Wrapf(err, "failed to delete the load balancer %q", *lb.Name)
+				}
+				deletedLBList[*lb.ID] = false
+			}
+			resourceLogger.WithField("name", *lb.Name).Info("load balancer deletetion triggered")
```
dharaneeshvrd, line 72: This line should also be inside if or else it would print for all the LB even if deletion not attempted.

Amulyam24: yes, missed it. Thanks for pointing it out.

internal/ibmcloud-janitor/resources/vpc_lbs.go
```diff
@@ -0,0 +1,118 @@
...
+func checkLBs(list map[string]bool, client *IBMVPCClient) error {
+	timeout := 0 * time.Minute
+	for timeout < lbTimeout {
+		for id, lbDeleted := range list {
+			lbID := id
+			if !lbDeleted {
+				lb, _, err := client.GetLoadBalancer(&vpcv1.GetLoadBalancerOptions{ID: &lbID})
+				if err != nil && strings.Contains(err.Error(), "cannot be found") {
+					list[id] = true
+					continue
```
dharaneeshvrd, line 104: This continue is not required

internal/ibmcloud-janitor/resources/vpc_lbs.go
```diff
@@ -0,0 +1,118 @@
...
+				} else if *lb.ProvisioningStatus == "delete_pending" {
+					break
+				} else {
+					return err
+				}
+			}
+		}
+		resourceLogger.Info("load balancer deletion is pending, checking again in a minute")
```
dharaneeshvrd, line 112: Need to have a check here for to validate the LBs are processed or not before sleeping. 
Even though all the LB's deleted, it would sleep and loop till timeout becomes greater than lbTimeout.
Need to refactor this logic.

It's better if we can add some unit test cases.

dharaneeshvrd: @Amulyam24 We can use https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait@v0.25.2#PollImmediate this func to query something on a interval with a timeout.

I think this suits our requirement. Can you please take a look at this?

e.g usage https://github.com/openshift/hypershift/blob/main/cmd/infra/powervs/create.go#L727

Amulyam24: @dharaneeshvrd, made changes as per your suggestion. Tested it locally too, and it works fine. The LBs deletion took a little less than 2 minutes. I have added a polling interval of 30 secs and a timeout as 4 minutes. PTAL.

dharaneeshvrd: Ok the new logic looks good to me.

internal/ibmcloud-janitor/resources/vpc_lbs.go
```diff
@@ -0,0 +1,114 @@
...
+			resourceLogger.Info("there are no available load balancers to delete")
+			return true, "", nil
+		}
+
+		for _, lb := range loadBalancers.LoadBalancers {
+			if *lb.ResourceGroup.ID == client.ResourceGroupID {
+				if _, err := client.DeleteLoadBalancer(&vpcv1.DeleteLoadBalancerOptions{
+					ID: lb.ID,
+				}); err != nil {
+					return false, "", errors.Wrapf(err, "failed to delete the load balancer %q", *lb.Name)
```
dharaneeshvrd, line 69: Do we need to come out of the loop before attempting to delete the remaining load balancers since one load balancer's delete API thrown an error?
IMO since it involves cleaning up all the load balancers, I would suggest to attempt the remaining load balancer deletion. And errors can be logged.
Same applies while monitoring the deletion as well, you can loop till end and any load balancer got an error while monitoring it can be logged.

Amulyam24: Addressed it, PTAL.

internal/ibmcloud-janitor/resources/vpc_lbs.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	if err = pagingHelper(f); err != nil {
+		errs = append(errs, errors.Wrapf(err, "failed to delete the load balancers"))
+	}
+
+	// check if the LBs are properly deleted
+	if err := checkLBs(deletedLBList, client, errs); err != nil {
+		errs = append(errs, errors.Wrapf(err, "failed to verify the deletion of load balancers"))
+	}
```
dharaneeshvrd, line 94:
```suggestion
	checkLBs(deletedLBList, client, errs)
```

internal/ibmcloud-janitor/resources/vpc_lbs.go
```diff
@@ -0,0 +1,123 @@
...
+
+	if len(errs) > 0 {
+		return kerrors.NewAggregate(errs)
+	}
+
+	resourceLogger.Info("Successfully deleted the load balancers")
+	return nil
+}
+
+func checkLBs(list []string, client *IBMVPCClient, errs []error) error {
```
dharaneeshvrd, line 104:
```suggestion
func checkLBs(list []string, client *IBMVPCClient, errs *[]error) {
```

internal/ibmcloud-janitor/resources/vpc_lbs.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	for _, lb := range list {
+		if err := check(lb); err != nil {
+			resourceLogger.WithField("ID", lb).Error("failed to check the deletion of load balancer")
+			errs = append(errs, err)
+		}
+	}
+
+	return kerrors.NewAggregate(errs)
```
dharaneeshvrd, line 122:
```suggestion
```

internal/ibmcloud-janitor/resources/vpc_lbs.go
```diff
@@ -0,0 +1,123 @@
...
+				return true, nil
+			}
+			return false, err
+		})
+	}
+
+	for _, lb := range list {
+		if err := check(lb); err != nil {
+			resourceLogger.WithField("ID", lb).Error("failed to check the deletion of load balancer")
+			errs = append(errs, err)
```
dharaneeshvrd, line 118:
```suggestion
			*errs = append(*errs, err)
```

internal/ibmcloud-janitor/resources/vpc_lbs.go
dharaneeshvrd: If we don't pass by ref, then it's kind of appending errs with aggregated errs.

Amulyam24: Addressed it. PTAL

dharaneeshvrd:
LGTM
PTAL @mkumatag

mkumatag (approved):
/lgtm

PR #162, opened 2023-02-02
Flag guard the disassociate public IP feature

youyangl:
Also, ignore the public IP description errors since error when running DescribeNetworkInterfacesPages should not block the rest part of internet gateway clean up logic

### DIFF
--- a/aws-janitor/resources/internet_gateways.go
+++ b/aws-janitor/resources/internet_gateways.go
@@ -66,9 +66,6 @@ func (InternetGateways) MarkAndSweep(opts Options, set *Set) error {
 			continue
 		}
 		logger.Warningf("%s: deleting %T: %s (%s)", i.ARN(), ig, i.ID, tags[NameTagKey])
-		if opts.DryRun {
-			continue
-		}
 
 		isDefault := false
 		for _, att := range ig.Attachments {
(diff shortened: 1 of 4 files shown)


aws-janitor/resources/internet_gateways.go
```diff
@@ -77,36 +74,46 @@ func (InternetGateways) MarkAndSweep(opts Options, set *Set) error {
...
+							Values: []*string{aws.String(*att.VpcId)},
+						},
 					},
-				},
-			}, pageFunc); err != nil {
-				return err
-			}
+				}, pageFunc); err != nil {
+					logger.Warningf("fail to get public IP for vpc %s: %v", *att.VpcId, err)
+				}
```
youyangl, line 96: We log the error but not return, the reason is that some times, the DescribeNetworkInterfacesPages returns not-nil error, making the rest resources clean up logic terminates in the middle.

aws-janitor/resources/internet_gateways.go
```diff
@@ -77,36 +74,46 @@ func (InternetGateways) MarkAndSweep(opts Options, set *Set) error {
...
 
-			// According to https://docs.aws.amazon.com/vpc/latest/userguide/VPC_Internet_Gateway.html#detach-igw
-			// Before detaching the internet gateway, we must dissassociate elastic IPs first.
-			for _, publicIP := range publicIPsToRelease {
-				disassociateReq := &ec2.DisassociateAddressInput{
-					PublicIp: aws.String(*publicIP),
+				logger.Warningf("%s: disassociating public IPs for vpc: %s", i.ARN(), *att.VpcId)
+				if opts.DryRun {
```
youyangl, line 99: Move dry run check here to better testing the DescribeNetworkInterfacesPages logic in dry run env

aws-janitor/resources/internet_gateways.go
```diff
@@ -77,36 +74,46 @@ func (InternetGateways) MarkAndSweep(opts Options, set *Set) error {
 				break
 			}
 
-			var publicIPsToRelease []*string
+			if opts.DisassociatePublicIP {
```
chemikadze, line 77: I am really not sure we need to keep this code - addresses are supposed to be disassociated when higher level resources are deleted, e.g. VMs or ELBs. If there would be addresses that warrant explicit deletion, it would likely be part of the code base, and earlier observed  IGW deletion issues are likely due to untagged ELBs that end up outliving IGWs and holding addresses associated.

We probably could roll back commit that introduced this and get rid of additional complexity.

aws-janitor/resources/internet_gateways.go
```diff
@@ -77,36 +74,46 @@ func (InternetGateways) MarkAndSweep(opts Options, set *Set) error {
...
+				var publicIPsToRelease []*string
 
-			pageFunc := func(page *ec2.DescribeNetworkInterfacesOutput, _ bool) bool {
-				for _, eni := range page.NetworkInterfaces {
-					publicIPsToRelease = append(publicIPsToRelease, eni.Association.PublicIp)
+				pageFunc := func(page *ec2.DescribeNetworkInterfacesOutput, _ bool) bool {
+					for _, eni := range page.NetworkInterfaces {
+						publicIPsToRelease = append(publicIPsToRelease, eni.Association.PublicIp)
```
chemikadze, line 82: I think that's the place where panic was happening originally: https://github.com/kubernetes-sigs/boskos/pull/158/files#r1083050091

Do we need to check that Association is not nil here?

aws-janitor/resources/internet_gateways.go
youyangl: I see, I can close this and send the revert one: https://github.com/kubernetes-sigs/boskos/pull/163

PR #164, opened 2023-02-07
fix(docs): update path to boskos-resources.yaml example

bweston92:


bweston92:
/remove-lifecycle stale

Should be a quick review.

PR #165, opened 2023-02-09
Delete IAM policies when detaching them from IAM roles

youyangl:
Currently, when detaching IAM policies from IAM roles, we don't delete them and leave them orphaned.
In this change, when detaching IAM policies from IAM roles, we also check the AttachmentCount of those policies. If they no longer have any roles attached to, we delete them.

### DIFF
--- a/aws-janitor/resources/iam_roles.go
+++ b/aws-janitor/resources/iam_roles.go
@@ -48,6 +48,14 @@ func fetchRoleAndTags(svc *iam.IAM, roleName *string) (*iam.Role, Tags, error) {
 	return role, tags, nil
 }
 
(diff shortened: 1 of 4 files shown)


aws-janitor/resources/iam_roles.go
```diff
@@ -91,9 +91,7 @@ func (IAMRoles) MarkAndSweep(opts Options, set *Set) error {
 					continue
 				}
 				logger.Warningf("%s: deleting %T: %s", l.ARN(), role, l.roleName)
-				if !opts.DryRun {
```
youyangl, line 94: Move dry run check to function: delete to better debugging in dry run mode.

PR #166, opened 2023-03-20
Work in federated / assume role scenarios"

dims:
Use sts `GetCallerIdentity` when `GetUser` fails... tested with the following on a federation system that uses assume role

```
aws-janitor \
  --ttl=0 \
  --region=us-west-2 \
  --dry-run=false \
  --log-level=debug \
  --exclude-tags=isengard \
  --enable-target-group-clean \
  --enable-vpc-endpoints-clean \
  --enable-dns-zone-clean \
  --path="s3://aws-janitor/cleanup"
```

without this patch, we were getting:
```
An error occurred (ValidationError) when calling the GetUser operation:
Must specify userName when calling with non-User credentials
```


richardcase (commented):
Seems reasonable to me

/lgtm

PR #169, opened 2023-05-19
Update gcloud version to the latest for the janitor image

hemendrateli:
gcloud behaviour has been changed recently for registering a cluster in a fleet. Now default behaviour is to register in cluster's region and not global. Previous gcloud versions still try to un-register from global location only and are not forward compatible.

### DIFF
--- a/images/janitor/Dockerfile
+++ b/images/janitor/Dockerfile
@@ -35,7 +35,7 @@ RUN make build
 ARG cmd
 RUN make "${cmd}"
 
-FROM gcr.io/google.com/cloudsdktool/cloud-sdk:375.0.0-slim
+FROM gcr.io/google.com/cloudsdktool/cloud-sdk:431.0.0-slim
 COPY ./cmd/janitor/gcp_janitor.py /bin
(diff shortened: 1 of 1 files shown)


hemendrateli:
Can someone please review this and if this looks good then please merge this.

chemikadze:
> Previous gcloud versions still try to un-register from global location only and are not forward compatible.

Do we need to modify janitor to take care of both regional AND global memberships?

hemendrateli:
> > Previous gcloud versions still try to un-register from global location only and are not forward compatible.
> 
> Do we need to modify janitor to take care of both regional AND global memberships?

We should not need to as "The registration behaviour in gcloud was changed to register gke clusters as regional memberships in gcloud version 422.0.0.  The membership region is selected to match the region of the gke cluster.  This is a nonbreaking change for gcloud because the gcloud membership commands were all changed to be backwards compatible with these changes"

krzyzacy (approved):

krzyzacy (approved):

PR #173, opened 2023-09-11
Remove password from client

upodroid:
This change eliminates the need for boskos client to rely on k/test-infra. 

Boskos shouldn't be exposed to the internet and definitely doesn't require user/password authentication.

I also need to revert a change in kubetest2 that I did simplify the deps being used.

```
 mahamed  REDACTED  ~  Desktop  …  kops  tests  e2e   fix-e2e-testing  2✎  3+  $  go run kubetest2-kops/main.go  --help
# k8s.io/kops/tests/e2e/kubetest2-kops/deployer
kubetest2-kops/deployer/common.go:94:15: cannot use boskosClient (variable of type *"sigs.k8s.io/kubetest2/internal/boskos/client".Client) as *"sigs.k8s.io/boskos/client".Client value in assignment
kubetest2-kops/deployer/common.go:97:5: cannot use d.boskos (variable of type *"sigs.k8s.io/boskos/client".Client) as *"sigs.k8s.io/kubetest2/internal/boskos/client".Client value in argument to boskos.Acquire
kubetest2-kops/deployer/down.go:64:4: cannot use d.boskos (variable of type *"sigs.k8s.io/boskos/client".Client) as *"sigs.k8s.io/kubetest2/internal/boskos/client".Client value in argument to boskos.Release
```


cmd/boskosctl/boskosctl.go
```diff
@@ -100,8 +98,8 @@ scripts with a simple interface.`,
 		Args: cobra.NoArgs,
 	}
 	root.PersistentFlags().StringVar(&options.serverURL, "server-url", "", "URL of the Boskos server")
-	root.PersistentFlags().StringVar(&options.username, "username", "", "Username used to access the Boskos server")
-	root.PersistentFlags().StringVar(&options.passwordFile, "password-file", "", "The path to password file used to access the Boskos server")
+	root.PersistentFlags().MarkDeprecated("username", "this flag has been deprecated")
+	root.PersistentFlags().MarkDeprecated("password-file", "this flag has been deprecated")
```
BenTheElder, line 102: not respected at all is more than deprecated. why even leave the flag?

BenTheElder (commented):
it's not clear why removing password is related to decoupling the two repos

cmd/boskosctl/boskosctl.go
upodroid: If someone was using it in their deployments, boskos would still start instead of crashlooping.

Happy to get rid of the flags properly

upodroid:
> it's not clear why removing password is related to decoupling the two repos

Using `sigs.k8s.io/boskos/client` imports k/test-infra which is required for the password feature. This feature has no functional value for those of us who run Prow clusters as boskos is an internal API that shouldn't require basic auth

upodroid:
bump, can I get this approved please?

stevekuznetsov:
I believe there are users of this feature out in the wild, not sure we can break the client constructor like this. I imagine even in-cluster clients may choose to use password auth.

PR #181, opened 2024-01-12
gcp_janitor.py: cleanup security-policies

sawsa307:
* Add security-policies to the list of resources for cleanup.
* Cleanup should be done for both regional and glocal policies.
* Security policies have dependencies on backend services, target pools, target instances, and instances. Thus, its cleanup should be after the cleanups of these resources.


sawsa307:
Tested locally with internal prow projects.

krzyzacy (approved):
/lgtm

PR #183, opened 2024-01-24
ssl-policies should be cleaned up after target-https-proxies.

sawsa307:
Based on https://cloud.google.com/load-balancing/docs/use-ssl-policies#delete-policy, ssl-policies should be cleaned up after target-https-proxies.

/assign @krzyzacy


krzyzacy (approved):
/lgtm

sawsa307:
Hi @krzyzacy Could you also lgtm this one? I think it should not have any issue now. Thank you!

PR #189, opened 2024-03-06
[gcp_janitor.py] Support regional instance-templates

jprzychodzen:
It's possible to create a regional instance template in GCP, however Janitor is unable to delete it.

Without this change, url to regional instance-template is malformed, for example `projects/jprzychodzen/global/instanceTemplates/templateFoo` instead of
`projects/jprzychodzen/us-central1/instanceTemplates/templateFoo`.

Tested manually with regional and global instanceTemplates.

Sample command to break janitor now:

```
gcloud compute instance-templates  create foo --project bar --instance-template-region us-central1
```


krzyzacy (approved):
/lgtm
/approve

PR #194, opened 2024-07-11
[gcp_janitor.py] Add removal of internal-ranges

jprzychodzen:
If network resource is left in the project, it's possible that we will leak internal-ranges object.

This change adds internal-ranges as an additional resource to deletion


krzyzacy (approved):
/ok-to-test

PR #200, opened 2024-07-31
Update to debian12 to get latest glibc

dims:
We reverted the previous image because of GLIBC issue https://github.com/kubernetes/test-infra/pull/33184

let's switch to newer debian which is used in k/release and elsewhere:
https://cs.k8s.io/?q=gcr.io%2Fdistroless%2Fbase-debian1&i=nope&files=&excludeFiles=&repos=


BenTheElder (commented):
/lgtm
🤞

PR #201, opened 2024-08-01
Use default region if LocationConstraint not found from s3 bucket

dims:
Failure in:
- https://storage.googleapis.com/kubernetes-jenkins/logs/maintenance-ci-shared-aws-janitor/1818809423779336192/build-log.txt
- https://storage.googleapis.com/kubernetes-jenkins/logs/ci-aws-ec2-janitor/1818809392036843520/build-log.txt

Because i missed some logic when converting from v1 to v2, specifically:
https://github.com/kubernetes-sigs/boskos/blame/210745d604c7bcd375251640e39b076a3467cb43/aws-janitor/s3/s3.go#L53-L56

### DIFF
--- a/aws-janitor/s3/s3.go
+++ b/aws-janitor/s3/s3.go
@@ -51,7 +51,10 @@ func GetPath(cfg *aws2.Config, s string) (*Path, error) {
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


tzneal (approved):
/lgtm

PR #213, opened 2025-01-03
Update Golang to `v1.23`

danilo-gemoli:
As per title, bump Golang to `v1.23.4` and also `golangci-lint`.


danilo-gemoli:
We need https://github.com/kubernetes/test-infra/pull/34066

PR #215, opened 2025-02-14
Added podman as an alternative engine for builds

hector-vido:
I leave the message of `which` just to made clear for someone that is observing the output that some kind of verification was happening.

We are preparing to build `boskos` images in our side - we can help the upstream or not, we wish but this depends on various discussions that are already happening.

With this simple modification `make images` can run with `podman` too, without messing what is already working.


images/build.sh
```diff
@@ -30,7 +30,18 @@ if [[ -z "${DOCKER_TAG:-}" ]]; then
     exit 1
 fi
 
-docker_build() {
+if [[ -z "${CONTAINER_ENGINE:-}" ]]; then
+    if  [[ -x `which docker` ]]; then
+        CONTAINER_ENGINE=docker
+    elif [[ -x `which podman` ]]; then
+        CONTAINER_ENGINE=podman
+    else
```
krzyzacy, line 38: fix indent here

also let's leave the default unchanged

hector-vido: I fixed the indent.
But what do you mean about "leave the default unchanged"?
Since docker is the first verification, it will become the default if the variable is unset.

krzyzacy (approved):

Issue #217, opened 2025-02-25
Build and publish Boskos images for ppc64le


Amulyam24:
This issue is to work on building and publish [images](https://github.com/kubernetes-sigs/boskos/blob/1230f5bd0accb7f5ff54d6e616d3da80a065c6d3/images/cloudbuild.yaml#L40) for ppc64le architecture.
All the images are not required, start with publishing

- [ ] boskos
- [ ] reaper
- [ ] ibmcloud-janitor-boskos

PR #219, opened 2025-03-05
Build and push multi-arch images for platforms amd64,arm64 and ppc64le

Amulyam24:
This PR adds changes for enabling multi arch build for boskos images for platforms x86_64 and ppc64le.

Issue: #217


images/build.sh
```diff
@@ -41,16 +43,35 @@ if [[ -z "${CONTAINER_ENGINE:-}" ]]; then
...
     fi
+
+    if [[ "${enable_buildx}" == 1 ]]; then
+        if [[ "${CONTAINER_ENGINE}" == "podman" ]]; then
+            echo "Podman is not preferred when ENABLE_BUILDX is set, terminating the build!"
+            exit 1 
+        else
+            ensure_buildx
+            echo "ENABLE_BUILDX is set! Building multi-arch image for '${cmd}' for platforms '${multi_arch_platforms}'" >&2
+            container_engine_build="${CONTAINER_ENGINE} buildx build --builder multiarch --platform=${multi_arch_platforms} --output type=registry"
```
Amulyam24, line 69: Added `--output type=registry` since `--output type=docker` or `--load` is not supported which stores the built images on local when two platforms are build simultaneously and errors with
`ERROR: docker exporter does not currently support exporting manifest lists`

If this change is okay, should the multi arch images be removed from the list here?https://github.com/kubernetes-sigs/boskos/blob/master/images/cloudbuild.yaml#L40

upodroid: Can we build multi arch by default? build amd64,arm64 and ppc64le

Amulyam24: @upodroid, added changes, PTAL.

upodroid (approved):
LGTM

PR #220, opened 2025-03-16
use a newer version of buildx

upodroid:
#219 failed because the buildx version is too old, bumping to latest


Amulyam24 (commented):
/lgtm

PR #231, opened 2025-06-02
Enhance the flow to skip PowerVS resource clean up using new state

Amulyam24:
This PR adds a feature to set a resource state to noSchedule when a tag is added to a PowerVS workspace. This will avoid it from being used for jobs or by janitor for cleanup until the tag is present.

Also, added more log statements and refactored the main to remove check for linter nestif.

### DIFF
--- a/cmd/ibmcloud-janitor-boskos/main.go
+++ b/cmd/ibmcloud-janitor-boskos/main.go
@@ -43,57 +43,107 @@ var (
 )
 
 const (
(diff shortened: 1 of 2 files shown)


Amulyam24:
@mkumatag, addressed the review comments by changing function names, using a ticker instead of time.Sleep and testing with multiple resources in `noSchedule` state.

mkumatag (approved):
/lgtm

Issue #218, opened 2025-03-03
Create releases for boskos


hector-vido:
Actually there is no releases for boskos, we just use the latest available version in the repos or build ourselves using the `main` branch.

Since Google Cloud Registry is being shutdown in March 2025, is a good think to create releases so we can promote its images in the same way as the other projects.

@jmguzik @krzyzacy

richackard:
I might be able to help with this one

PR #233, opened 2025-07-01
Proposed a release process for the Boskos project

richackard:
Re: Issue #218 

This PR proposes a lightweight release process for the Boskos project.

The script in this PR helps search and fetch the container image hashes for the images built from a given commit.

Note that the `registry.k8s.io/images/k8s-staging-boskos/images.yaml` doesn't exist in the `kubernetes/k8s.io` repo just yet because we need to have the first version tag (probably `v0.0.1` to commit `e9e5322`) in this repo before we map that version tag to the first set of images in the image promoter.

### DIFF
--- a/hack/check_images.sh
+++ b/hack/check_images.sh
@@ -1,0 +1,68 @@
+#!/usr/bin/env bash
+# Copyright 2025 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+REPO="gcr.io/k8s-staging-boskos"
+
+# Check if a commit SHA was provided as a command-line argument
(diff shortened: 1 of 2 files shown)


hack/check_images.sh
```diff
@@ -0,0 +1,68 @@
...
+if [ -z "$1" ]; then
+  echo "Usage: $0 <last_7_chars_commit_sha>"
+  echo "Please provide the last 7 chars of the commit hash to search for in the image tags."
+  exit 1
+fi
+
+COMMIT_SHA="$1"
+
+# Get a list of all images in the specified repository
+images=$(gcloud container images list --repository="$REPO" --format="get(name)")
```
BenTheElder, line 32: Hmm, what if there are random stray images?
I feel like we should identify them by some other means (also this script will only work with GCR/AR and not other repos because of this particular line)

hack/check_images.sh
```diff
@@ -0,0 +1,68 @@
...
+
+echo "Searching for tags containing '$COMMIT_SHA' in repository '$REPO'..."
+echo "---"
+
+matched_count=0
+
+printf "%-70s %s\n" "IMAGE_NAME" "MATCHED_IMAGE_SHA256"
+for image in $images; do
+  # Find the images whose tags match the commit hash
+  digest=$(gcloud container images list-tags "$image" --filter="tags~$COMMIT_SHA" --format="get(digest)")
```
BenTheElder, line 50: we can use `crane` as a portable option that isn't GCP specific, though let's not bother unless we replace the other gcloud dependencies

BenTheElder (commented):
LGTM, would like input from the current maintainers though (and also we may need to update that list with some more active contributors ...)

cc @kubernetes-sigs/boskos-maintainers / @kubernetes-sigs/boskos-admins

PR #235, opened 2025-08-21
Add support for cleaning ECR images in aws-janitor

ConnorJC3:
## Summary

Adds a new janitor feature to clean ECR container images from repos (does not delete the repos themselves). There is a limit of 10k images per repository, and every few months our CI hits this limit and we have to manually clean out the repo.

This janitor is disabled by default and must be enabled by CLI flag (1) to avoid breaking anyone's existing workflow and (2) because it is not possible to tag ECR repos, so it is not possible to skip repos with this cleaner, it will clean all ECR repos.

Also had to bump some AWS SDK dependencies to get this to work.

## Testing

Dry run test:
```
$ go run ./cmd/aws-janitor --region us-west-2 --enable-ecr-images-clean=true --all=true --dry-run=true |& jq -r '.msg'
Regions: us-west-2
arn:aws:ecr:us-west-2:368597081700:repository/csi-node-driver-registrar: deleting 5 images
arn:aws:ecr:us-west-2:368597081700:repository/csi-resizer: deleting 3 images
arn:aws:ecr:us-west-2:368597081700:repository/livenessprobe: deleting 5 images
arn:aws:ecr:us-west-2:368597081700:repository/csi-attacher: deleting 3 images
arn:aws:ecr:us-west-2:368597081700:repository/snapshot-controller: deleting 3 images
arn:aws:ecr:us-west-2:368597081700:repository/csi-provisioner: deleting 3 images
arn:aws:ecr:us-west-2:368597081700:repository/csi-snapshotter: deleting 3 images
arn:aws:ecr:us-west-2:368597081700:repository/aws-ebs-csi-driver: deleting 2 images
```

Real delete test:
```
$ go run ./cmd/aws-janitor --region us-west-2 --enable-ecr-images-clean=true --all=true |& jq -r '.msg'               
Regions: us-west-2
arn:aws:ecr:us-west-2:368597081700:repository/csi-node-driver-registrar: deleting 5 images
arn:aws:ecr:us-west-2:368597081700:repository/csi-resizer: deleting 3 images
arn:aws:ecr:us-west-2:368597081700:repository/livenessprobe: deleting 5 images
arn:aws:ecr:us-west-2:368597081700:repository/csi-attacher: deleting 3 images
arn:aws:ecr:us-west-2:368597081700:repository/snapshot-controller: deleting 3 images
arn:aws:ecr:us-west-2:368597081700:repository/csi-provisioner: deleting 3 images
arn:aws:ecr:us-west-2:368597081700:repository/csi-snapshotter: deleting 3 images
arn:aws:ecr:us-west-2:368597081700:repository/aws-ebs-csi-driver: deleting 2 images
$ aws ecr list-images --repository-name aws-ebs-csi-driver
{
    "imageIds": []
}
```

### DIFF
--- a/aws-janitor/resources/ecr.go
+++ b/aws-janitor/resources/ecr.go
@@ -1,0 +1,246 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package resources
+
+import (
+	"context"
+	"fmt"
+	"slices"
+	"time"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/ecr"
+	"github.com/aws/aws-sdk-go-v2/service/ecr/types"
+
+	"github.com/sirupsen/logrus"
+)
+
+type ecrClient interface {
+	DescribeRepositories(ctx context.Context, params *ecr.DescribeRepositoriesInput, optFns ...func(*ecr.Options)) (*ecr.DescribeRepositoriesOutput, error)
+	DescribeImages(ctx context.Context, params *ecr.DescribeImagesInput, optFns ...func(*ecr.Options)) (*ecr.DescribeImagesOutput, error)
+	BatchDeleteImage(ctx context.Context, params *ecr.BatchDeleteImageInput, optFns ...func(*ecr.Options)) (*ecr.BatchDeleteImageOutput, error)
+}
+
+type ContainerImages struct{}
+
+func (ContainerImages) MarkAndSweep(opts Options, set *Set) error {
+	logger := logrus.WithField("options", opts)
+	if len(opts.CleanEcrRepositories) == 0 {
+		logger.Info("No ECR reposotories to clean provided, skipping ECR")
+		return nil
+	}
+
+	svc := ecr.NewFromConfig(*opts.Config, func(opt *ecr.Options) {
+		opt.Region = opts.Region
+	})
+	return markAndSweepWithClient(svc, opts, set)
+}
+
+//nolint:gocognit // Just over the complexity threshold and splitting would make it less readable
+func markAndSweepWithClient(svc ecrClient, opts Options, set *Set) error {
+	logger := logrus.WithField("options", opts)
+	inp := &ecr.DescribeRepositoriesInput{
+		RegistryId: aws.String(opts.Account),
+	}
+
+	// DescribeImages requires a repository name, so we first must crawl all repositories
+	err := DescribeRepositoriesPages(svc, inp, func(repos *ecr.DescribeRepositoriesOutput) error {
+		for _, repo := range repos.Repositories {
+			if !slices.Contains(opts.CleanEcrRepositories, *repo.RepositoryName) {
+				continue
+			}
+
+			// BatchDeleteImage can only be called per-repo, so describe all the images in a repo, delete them, and repeat
+			var toDelete []*image
+			imageInp := &ecr.DescribeImagesInput{
+				RegistryId:     repo.RegistryId,
+				RepositoryName: repo.RepositoryName,
+			}
+
+			imagesErr := DescribeImagesPages(svc, imageInp, func(images *ecr.DescribeImagesOutput) error {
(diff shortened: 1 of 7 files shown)


aws-janitor/resources/ecr.go
```diff
@@ -0,0 +1,218 @@
...
+			}
+
+			err := DescribeImagesPages(svc, imageInp, func(images *ecr.DescribeImagesOutput) error {
+				for _, ecrImage := range images.ImageDetails {
+					i := image{
+						Registry:   *ecrImage.RegistryId,
+						Region:     opts.Region,
+						Repository: *ecrImage.RepositoryName,
+						Digest:     *ecrImage.ImageDigest,
+					}
```
torredil, line 81: Do we need nil checks before dereferencing these pointers? same for L134-138 below.

aws-janitor/resources/ecr.go
```diff
@@ -0,0 +1,218 @@
...
+	err := DescribeRepositoriesPages(svc, inp, func(repos *ecr.DescribeRepositoriesOutput) error {
+		for _, repo := range repos.Repositories {
+			// BatchDeleteImage can only be called per-repo, so describe all the images in a repo, delete them, and repeat
+			var toDelete []*image
+			imageInp := &ecr.DescribeImagesInput{
+				RegistryId:     repo.RegistryId,
+				RepositoryName: repo.RepositoryName,
+			}
+
+			err := DescribeImagesPages(svc, imageInp, func(images *ecr.DescribeImagesOutput) error {
```
torredil: Suggest renaming this since its shadowing the above err:
```
err := DescribeRepositoriesPages
```

aws-janitor/resources/ecr.go
```diff
@@ -0,0 +1,218 @@
...
+				// If we've reached 100 images, or this is the last item in the toDelete slice, make the call to the AWS API
+				if len(deleteReq.ImageIds) == 100 || n == len(toDelete)-1 {
+					logger.Warningf("%s: deleting %d images", *repo.RepositoryArn, len(deleteReq.ImageIds))
+					if !opts.DryRun {
+						result, err := svc.BatchDeleteImage(context.TODO(), deleteReq)
+
+						if err != nil {
+							logger.Warningf("%s: delete API call failed: %v", *repo.RepositoryArn, err)
+						} else if len(result.Failures) > 0 {
+							logger.Warningf("%s: some images failed to delete: %T", *repo.RepositoryArn, result.Failures)
```
torredil, line 115: I think we need to reset the slice. Say theres 200 images to delete, after the first batch of 100 gets deleted the second batch will still have the previous items, its total size will be 200 but it should only have 101-200.

ConnorJC3: Ah yes, you're totally right we do - fixed in next revision.

aws-janitor/resources/ecr.go
ConnorJC3: Fixed in next rev.

aws-janitor/resources/ecr.go
ConnorJC3: I don't believe so - the response object should always contain them. I could add a check if we wanted to be ultra-safe, but skimming the rest of the janitor we don't seem to do such checks.

torredil:
/lgtm

cc: @dims

dims:
@ConnorJC3 i understand and agree that false by default is helpful. but `so it is not possible to skip repos with this cleaner, it will clean all ECR repos.` can still lead to potentially bad outcomes, can you please add an allow list? only those specified in the allow list will be cleaned? (empty allow list means, don't do anything! so let's print a helpful message)

ConnorJC3:
@dims good idea, newest rev switches the flag from a binary true/false to a list of repos to clean

ConnorJC3:
/assign stevekuznetsov

Assigning @stevekuznetsov as `k8s-ci-robot` suggested, when you get a chance can you take a look? Thanks!

krzyzacy:
do you still need this PR?

/assign

ConnorJC3:
@krzyzacy yes, if you're willing to review/merge it please do - thanks! I was waiting for the holiday seasion to be over before pinging people again.

aws-janitor/resources/ecr.go
```diff
@@ -0,0 +1,231 @@
+/*
+Copyright 2025 The Kubernetes Authors.
```
krzyzacy: could you add some basic unittest coverage for the new method you've added? (I know the existing one doesn't have tests, but we probably should have some coverage)

ConnorJC3: I added some test cases that should test the majority of the functionality of `ListAll` and `MarkAndSweep`.

In order to do this, I had to refactor most of their logic into a helper, and leave only instantiating the ECR client in the main functions. This is because I need to be able to pass in a mock client for the tests, but cannot change the signature of those functions or the object wouldn't conform to the interface.

_(Also rebased the branch, but if you're looking at the GitHub diff the only changes are to `ecr.go` and the addition of `ecr_test.go`)_

krzyzacy (approved):
Thank you!

Issue #236, opened 2025-08-28
ibmcloud-janitor: avoid restart on failure with better error handling


Amulyam24:
When a resource such as VM/network fails to be cleaned up due to any issue, the janitor restarts continuously until the resource is cleaned up successfully. This task is to explore if there is a better way to handle this scenario.

Amulyam24:
fixed via #239

PR #239, opened 2025-09-18
Fix ibmcloud-janitor restart on cleanup failure issue

shwetha-s-poojary:
### **Issue and root cause:**
Janitor was restarting whenever a cleanup operation failed.The janitor process was using a return statement inside its infinite loop to handle cleanup failures. This caused the loop to exit gracefully whenever a cleanup error occurred, leading to premature termination of the janitor service.

### **Fix:**
Instead of returning the error, logged the error and continue to ensure the loop keeps running even after cleanup failures.

### **Test:**
Confirmed that cleanup errors are logged and the loop continues without triggering a restart.

Screenshot 2025-09-18 at 2 54 33 PM

### DIFF
--- a/cmd/ibmcloud-janitor-boskos/main.go
+++ b/cmd/ibmcloud-janitor-boskos/main.go
@@ -74,7 +74,8 @@ func run(boskos *boskosClient.Client) error {
 			}
 			options.Resource = res
 			if err := resources.CleanAll(options); err != nil {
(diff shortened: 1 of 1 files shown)


shwetha-s-poojary:
Fixes #236

Amulyam24 (approved):
@shwetha-s-poojary, Thank you for the fix!

/lgtm

PR #244, opened 2026-02-24
Protect Lustre e2e resources from gcp janitor

chrisThePattyEater:
What type of PR is this?
/kind feature

What this PR does / why we need it:
This pr introduces the Lustre resources generated during our e2e tests as preserved_names. This should prevent these resources from being cleaned up after our tests complete which will prevent the need to recreate later.

* VPC, Subnet and IP Ranges now have the lustre resources as preserved names
* preserved names now also accept regex
* Secondary ip ranges are handled by `clean_secondary_ip_ranges` which does not respect the preserved_names, it will attempt to clean up secondary ip ranges, and therefore needs its own manual skip.

### DIFF
--- a/cmd/janitor/gcp_janitor.py
+++ b/cmd/janitor/gcp_janitor.py
@@ -21,9 +21,14 @@ import collections
 import datetime
 import json
 import os
+import re
 import subprocess
 import sys
 import threading
+
+# Lustre CSI Constants
+LUSTRE_REGEX = 'lustre-.*'
+LUSTRE_MULTINIC_SUBNET = 'multinic-subnet'
 
 # A resource that need to be cleared.
 # Any names in preserved_names will never be deleted.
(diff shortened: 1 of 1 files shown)


krzyzacy:
could you link to an upstream issue? (is this breaking k8s CI?)

what if these network leaks, do you have a secondary process to clean up?

cmd/janitor/gcp_janitor.py
```diff
@@ -25,6 +25,11 @@
 import sys
 import threading
 
+# Lustre CSI Constants
+LUSTRE_PEERING_RANGE = 'lustre-peering-range'
```
tyuchn: Should the IP range name be `lustre-range`?
https://github.com/GoogleCloudPlatform/lustre-csi-driver/blob/main/test/k8s-integration/network.go#L29

Nit: Would it be better to use a regex name here, such as `lustre-*`? The networking resources appear to be hardcoded in the Lustre CSI driver test, which means any name changes in the CSI driver repo would invalid the janitor script

krzyzacy (approved):

PR #245, opened 2026-03-10
implement pruning workflow

rvanderp3:
** changes **
- build aws-janitor as image
- recognize preserve tag by default


rvanderp3:
opened this PR by accident

Issue #250, opened 2026-04-27
gcp_janitor.py fails to delete auto-generated peering routes during project cleanup


emerbe:
When running gcp_janitor.py to clean up resources in a GCP project (specifically with --hours 0 or for projects in "cleaning" state), the script fails when attempting to delete routes that were auto-generated by VPC Network Peerings.

GCP does not allow deleting these routes directly; they are managed by the lifecycle of the peering itself. Trying to delete them results in the following error:

```
ERROR: (gcloud.compute.routes.delete) Could not fetch resource:
 - The auto-generated peering route cannot be deleted.
```

PR #251, opened 2026-04-27
Janitor: Fix auto-generated peering route deletion error

emerbe:
GCP does not allow deleting routes that are auto-generated by VPC Network
Peerings directly. Attempting to do so results in a CalledProcessError.
These routes are automatically removed when the peering itself is deleted.
This commit fixes the issue by:
1. Adding a `clean_network_peerings` function to delete peerings before
   regular resource cleanup, which clears the associated routes.
2. Updating `validate_item` to skip routes that have a non-empty
   `nextHopPeering` field, ensuring we don't fail on routes from ignored
   networks.
3. Moving the special handling for protected networks and peering routes
   before the `skip_age_check` (formerly `clear_all`) return in
   `validate_item`.
4. Renaming the `clear_all` variable to `skip_age_check` across the
   file to better describe its function.
This ensures that project cleanup can complete successfully without
failing on impossible route deletions.

### DIFF
--- a/cmd/janitor/gcp_janitor.py
+++ b/cmd/janitor/gcp_janitor.py
@@ -239,14 +239,14 @@ def base_command(resource):
     return base
 
 
-def validate_item(item, age, resource, clear_all):
+def validate_item(item, age, resource, skip_age_check):
     """ Validate if an item need to be cleaned.
 
     Args:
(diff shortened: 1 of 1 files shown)


emerbe:
resolves: #250

krzyzacy (approved):

PR #256, opened 2026-05-21
refactor client, using separate module

pohly:
The goal is to make simpleclient self-contained and consumable in Kubernetes.
The main sigs.k8s.io/boskos package may depend on it, but not the other way around.
    
To achieve this, the common and storage package get moved into simpleclient.
The existing packages, including client, are kept to avoid breaking consumers
of the boskos package. This may be overly cautious, but who knows... One
benefit here is that consumers in the repo do not need to be updated.
    
Loading secrets from a file remains in client/client.go because it
has too many undesirable dependencies.

A replace statement ensures that within the repo the code is always consistent.
Consumers of the boskos package (if there are any) must keep sigs.k8s.io/boskos
and sigs.k8s.io/boskos/simpleclient at the same revision.


pohly:
This is an alternative to forking the client in https://github.com/kubernetes/kubernetes/pull/139129.

BenTheElder:
> This is an alternative to forking the client in https://github.com/kubernetes/kubernetes/pull/139129.

What does the k/k PR look like with this commit?

pohly:
I haven't tried using this in k/k yet, definitely something to do first before merging. My expectation is that the PR will then add boskos/simpleclient to vendor and nothing else, without having to create test/utils/boskos.

It might be useful to add my updated Acquire here, then it's also readily available:

https://github.com/pohly/kubernetes/blob/7d9912a1966e1a114998f1d4ce1350a9d7a8821a/test/utils/boskos/client/boskos.go

But it's small enough that a local copy in test/e2e_node would also be fine.

pohly:
Yes, that worked: https://github.com/kubernetes/kubernetes/pull/139129/changes/007e6c00021342984a9516b5621e354659ccb3c8

```console
$ git log --stat -n1 vendor
commit 007e6c00021342984a9516b5621e354659ccb3c8 (HEAD -> e2e-node-update-local, pohly/e2e-node-update-local)
Author: Patrick Ohly <patrick.ohly@intel.com>
Date:   Thu May 21 22:51:07 2026 +0200

    WIP: use https://github.com/kubernetes-sigs/boskos/pull/256

 vendor/modules.txt                                            |   6 +
 vendor/sigs.k8s.io/boskos/simpleclient/LICENSE                | 201 ++++++++++++++++++
 vendor/sigs.k8s.io/boskos/simpleclient/README.md              |  69 +++++++
 vendor/sigs.k8s.io/boskos/simpleclient/client.go              | 719 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 vendor/sigs.k8s.io/boskos/simpleclient/common/common.go       | 301 +++++++++++++++++++++++++++
 vendor/sigs.k8s.io/boskos/simpleclient/common/mason_config.go |  96 +++++++++
 vendor/sigs.k8s.io/boskos/simpleclient/storage/storage.go     |  99 +++++++++
```

pohly:
I kept `test/utils/boskos/client/boskos.go` with `Acquire` in that commit; that's not where it should remain, I just wanted a quick POC.

pohly:
`simpleclient.Client` looks a bit odd in the code which consumes the new module. If we are okay with a breaking Go API change in the sigs.k8s.io/boskos module, then I can rename `client` to `fileclient` and `simpleclient` back to `client`. That's what I did originally, but then callers of `NewClient` must be updated because the function signature is different, so I decided to avoid that.

upodroid:
> Loading secrets from a file remains in client/client.go because it
> has too many undesirable dependencies.

This is something we don't want to continue doing so let's remove it and see if we can do the change in place. I tried this a while back in #173

pohly:
This separate client module may have value, but right now I probably no longer need it myself: for https://github.com/kubernetes/kubernetes/pull/139129 I am going to keep the Boskos acquire in an enhanced kubetest2 node tester, see https://github.com/kubernetes-sigs/kubetest2/pull/338.

We can revisit this when the need arises again.

/close

PR #257, opened 2026-05-29
Delete Instance Templates and other resources based on VPC ID

sudharshanibm:
- Delete Instance Templates
- Delete Loadbalancer and Floating IPS
- Deletion of all resources based on VPC ID

### DIFF
--- a/internal/ibmcloud-janitor/resources/vpc_floating_ips.go
+++ b/internal/ibmcloud-janitor/resources/vpc_floating_ips.go
@@ -1,0 +1,119 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package resources
+
+import (
+	"context"
+	"strings"
+	"time"
+
+	"github.com/IBM/vpc-go-sdk/vpcv1"
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+var (
+	fipDeletionTimeout  = time.Minute
+	fipPollingInterval  = time.Second * 10
+	fipNotFoundPatterns = []string{"cannot be found", "not found"}
+)
+
+func deleteResourceGroupFloatingIPs(client *IBMVPCClient, resourceLogger *logrus.Entry) error {
+	fips, _, err := client.ListFloatingIps(&vpcv1.ListFloatingIpsOptions{
+		ResourceGroupID: &client.ResourceGroupID,
+	})
+	if err != nil {
+		return errors.Wrap(err, "failed to list the floating IPs")
+	}
+	for _, fip := range fips.FloatingIps {
+		if fip.ID == nil {
+			continue
+		}
+		if err := deleteFloatingIP(client, *fip.ID, resourceLogger); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func deleteFloatingIP(client *IBMVPCClient, id string, resourceLogger *logrus.Entry) error {
+	if err := triggerFloatingIPDeletion(client, id); err != nil {
+		return err
+	}
+	if err := waitForFloatingIPDeleted(client, id); err != nil {
+		return err
+	}
+	resourceLogger.WithField("id", id).Info("Successfully deleted the floating IP")
+	return nil
+}
+
+func triggerFloatingIPDeletion(client *IBMVPCClient, id string) error {
+	var lastErr error
+	err := wait.PollUntilContextTimeout(context.Background(), fipPollingInterval, fipDeletionTimeout, true, func(_ context.Context) (bool, error) {
+		_, err := client.DeleteFloatingIP(&vpcv1.DeleteFloatingIPOptions{
+			ID: &id,
+		})
+		if err == nil {
+			return true, nil
+		}
+		if isFloatingIPNotFound(err) {
+			return true, nil
+		}
+		lastErr = err
+		return false, nil
+	})
+	if err != nil {
+		if lastErr != nil {
+			err = lastErr
+		}
+		return errors.Wrapf(err, "failed to delete the floating IP %q", id)
+	}
+	return nil
+}
+
+func waitForFloatingIPDeleted(client *IBMVPCClient, id string) error {
+	var lastErr error
+	err := wait.PollUntilContextTimeout(context.Background(), fipPollingInterval, fipDeletionTimeout, true, func(_ context.Context) (bool, error) {
+		_, _, err := client.GetFloatingIP(&vpcv1.GetFloatingIPOptions{ID: &id})
+		if err == nil {
+			return false, nil
+		}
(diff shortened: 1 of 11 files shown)


internal/ibmcloud-janitor/resources/vpc_instances.go
```diff
@@ -48,13 +58,86 @@ func (VPCInstance) cleanup(options *CleanupOptions) error {
 	}
 
 	for _, ins := range instanceList.Instances {
+		if ins.ID == nil {
+			continue
+		}
+		instanceName := *ins.ID
```
Amulyam24: did you mean?  
```suggestion
		instanceID := *ins.ID
```
is this needed?

internal/ibmcloud-janitor/resources/vpc_floating_ips.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func deleteFloatingIP(client *IBMVPCClient, id string, resourceLogger *logrus.Entry) error {
+	var lastErr error
+	err := wait.PollUntilContextTimeout(context.Background(), fipPollingInterval, fipDeletionTimeout, true, func(_ context.Context) (bool, error) {
+		_, err := client.DeleteFloatingIP(&vpcv1.DeleteFloatingIPOptions{
+			ID: &id,
+		})
+		if err == nil {
+			return true, nil
```
Amulyam24, line 73: should this be false?

internal/ibmcloud-janitor/resources/vpc_instances.go
```diff
@@ -48,13 +58,86 @@ func (VPCInstance) cleanup(options *CleanupOptions) error {
...
+		}
+		lastErr = err
+		return false, nil
+	})
+	if err != nil {
+		if lastErr != nil {
+			return errors.Wrapf(lastErr, "timed out waiting for instance %q to be deleted", name)
+		}
+		return errors.Wrapf(err, "timed out waiting for instance %q to be deleted", name)
+	}
```
Amulyam24, line 133:
```suggestion
	if err != nil {
		if lastErr != nil {
			err = lastErr
		}
		return errors.Wrapf(err, "timed out waiting for instance %q to be deleted", name)
	}
```

Would this work?
applicable to all resource wait functions

internal/ibmcloud-janitor/resources/vpc_public_gateways.go
```diff
@@ -0,0 +1,81 @@
...
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+func deleteSubnetPublicGateway(client *IBMVPCClient, subnet vpcv1.Subnet, resourceLogger *logrus.Entry) error {
+	pg, _, err := client.GetSubnetPublicGateway(&vpcv1.GetSubnetPublicGatewayOptions{
+		ID: subnet.ID,
+	})
+	if pg == nil || err != nil {
```
Amulyam24: should we handle the `err` here?

internal/ibmcloud-janitor/resources/vpc_networks.go
```diff
@@ -51,52 +50,23 @@ func (VPCNetwork) cleanup(options *CleanupOptions) error {
...
-			// Skip bound FIPs if VPC ID is specified
-			continue
-		}
-		_, err = client.DeleteFloatingIP(&vpcv1.DeleteFloatingIPOptions{
-			ID: fip.ID,
-		})
-		if err != nil {
-			return errors.Wrapf(err, "failed to delete the floating IP %q", *fip.Name)
+	if client.VPCID != "" {
+		if err := deleteTargetVPCPublicGateways(client, resourceLogger); err != nil {
```
Amulyam24, line 64: any reason for adding this as we are already listing the subnets associated with the target VPC above?

internal/ibmcloud-janitor/resources/vpc_floating_ips.go
sudharshanibm: `true` is intentional here
polling condition returns `true` when deletion succeeds so retries stop immediately

Amulyam24: If there is no error and we don't retry after returning true here, is there a need to check if the FIP is deleted below? wasn't that the initial intention similar to `waitForInstanceDeleted`.

sudharshanibm: yeah got it, will add a `waitForFloatingIPDeleted() `check using `GetFloatingIP()` similar to the instance-template deletion flow

internal/ibmcloud-janitor/resources/vpc_public_gateways.go
sudharshanibm: updated the code now `404` is treated as no gateway attached
other API errors are returned

internal/ibmcloud-janitor/resources/vpc_networks.go
sudharshanibm: in the recent s390x conformance runs we noticed that when a job fails boskos sometimes leaves the public gateway undeleted and leaves them as stale resource
here the second pass catches orphan public gateways left in the target vpc after a partial cleanup

internal/ibmcloud-janitor/resources/vpc_instances.go
```diff
@@ -48,13 +58,87 @@ func (VPCInstance) cleanup(options *CleanupOptions) error {
 	}
 
 	for _, ins := range instanceList.Instances {
+		if ins.ID == nil {
+			continue
+		}
+		instanceID := *ins.ID
+		instanceName := instanceID
```
Amulyam24: maybe this could be removed?

Amulyam24 (commented):
Minor comment, LGTM

Thank you @sudharshanibm!

internal/ibmcloud-janitor/resources/vpc_instances.go
sudharshanibm: removed only the `instanceName := instanceID` fallback now. 
api calls still use `instanceID` and logs use the instance name
kept the nil check as a defensive guard

internal/ibmcloud-janitor/resources/vpc_instance_templates.go
```diff
@@ -0,0 +1,104 @@
...
+		if !instanceTemplateBelongsToVPC(template, targetVPCIDs) {
+			continue
+		}
+		_, err := client.DeleteInstanceTemplate(&vpcv1.DeleteInstanceTemplateOptions{
+			ID: template.ID,
+		})
+		if err != nil {
+			return errors.Wrapf(err, "failed to delete the instance template %q", *template.Name)
+		}
+		resourceLogger.WithFields(logrus.Fields{"name": template.Name}).Info("Successfully deleted the instance template")
```
Amulyam24: Should this be printing `*template.Name` instead?

Please cross check in other resources if we are printing address instead of value.

sudharshanibm: updated the `instance-template` one and cross checked all other resources
fixed the issue in `vpc_public_gateways.go `and `vpc_lbs.go` 
btw `serviceid.go` has the same pattern but it's already existing and unrelated to this PR
can i update that as well?

Amulyam24: > btw serviceid.go has the same pattern but it's already existing and unrelated to this PR

We can fix that in a follow-up PR, do you mind creating it?

Amulyam24 (approved):
/lgtm
/approve

Thank you @sudharshanibm!

internal/ibmcloud-janitor/resources/vpc_instance_templates.go
sudharshanibm: sure will create that

sudharshanibm: Fix for `serviceid.go` : #258

PR #259, opened 2026-07-09
gcp_janitor: clean up service attachments

supahero1:
Fixes resource leakage and infinite dirty -> janitor cycles

### DIFF
--- a/cmd/janitor/gcp_janitor.py
+++ b/cmd/janitor/gcp_janitor.py
@@ -69,6 +69,7 @@ RESOURCES_BY_API = {
         Resource('', 'compute', 'disks', None, 'zone', None, False, True, None),
         Resource('', 'compute', 'disks', None, 'region', None, False, True, None),
         Resource('', 'compute', 'firewall-rules', None, None, None, False, True, None),
+        Resource('', 'compute', 'service-attachments', None, 'region', None, False, True, None),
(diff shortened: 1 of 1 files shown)


cmd/janitor/gcp_janitor.py
```diff
@@ -69,6 +69,7 @@
         Resource('', 'compute', 'disks', None, 'zone', None, False, True, None),
         Resource('', 'compute', 'disks', None, 'region', None, False, True, None),
         Resource('', 'compute', 'firewall-rules', None, None, None, False, True, None),
+        Resource('', 'compute', 'service-attachments', None, 'region', None, False, True, None),
```
krzyzacy, line 72: does this object have gcloud version dependency? if so please also bump https://github.com/kubernetes-sigs/boskos/blob/master/images/janitor/Dockerfile#L38

krzyzacy: also does the order matter?

supahero1: Janior is using 568 whereas service attachments were added in 349 so all good here.

Yes, order matters, and I think it's in the right place, because it has to be deleted before forwarding rules, otherwise `resourceInUseByAnotherResource` would be thrown.

krzyzacy (approved):

PR #260, opened 2026-07-17
Add FSx for Lustre cleanup

torredil:
Adds FSx for Lustre to the AWS janitor. Currently, e2e runs in https://github.com/kubernetes-sigs/aws-fsx-csi-drive can leave filesystems behind, and cleanup in the repo only runs when presubmits run.

### DIFF
--- a/aws-janitor/resources/list.go
+++ b/aws-janitor/resources/list.go
@@ -86,6 +86,7 @@ var RegionalTypeList = []Type{
 	LaunchConfigurations{},
 	LaunchTemplates{},
 	Instances{},
+	FSxLustreFileSystems{},
 	VPCEndpoints{},
 	// Addresses
 	NetworkInterfaces{},
--- a/aws-janitor/resources/fsx.go
+++ b/aws-janitor/resources/fsx.go
@@ -1,0 +1,180 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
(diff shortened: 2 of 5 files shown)


torredil:
cc: @khoang98

khoang98:
Thank you for your help! @torredil

torredil:
Hello @krzyzacy can you give us a review on this one? thank you : )

krzyzacy (approved):

aws-janitor/resources/list.go
```diff
@@ -86,6 +86,7 @@ var RegionalTypeList = []Type{
 	LaunchConfigurations{},
 	LaunchTemplates{},
 	Instances{},
+	FSxLustreFileSystems{},
```
krzyzacy, line 89: I'm not AWS expert, does the order matter here? (do we need to document these somethere?)

krzyzacy:
/hold

minor question above

aws-janitor/resources/list.go
torredil: Yeah FSx should be cleaned up before the networking resources it uses. I think the comment [right above](https://github.com/torredil/boskos/blob/3563df0d3777e1551623cd6640266a5ab888f37b/aws-janitor/resources/list.go#L79) already covers this by saying the list is in dependency order, but let me know if we need a more detailed explanation in the source code.

PR #263, opened 2026-07-29
aws-janitor: continue sweeping remaining resource types on error

### LINKED ISSUE #262: aws-janitor aborts the entire sweep on the first error from any resource type
This issue tracks the following action item:

> The aws-janitor hard-fails the entire sweep when a single resource type hits AccessDenied (see https://github.com/kubernetes-sigs/boskos/pull/260). It would be more robust to log the error for that resource type and continue sweeping the rest.  - @dims

A failure in one resource type shouldn't take out cleanup for everything else. The janitor should log it, keep sweeping the remaining types, and then exit non-zero so the job reports the failure.

See https://github.com/kubernetes/k8s.io/issues/9790 for more context.

torredil:
**What this PR does / why we need it**:

`markAndSweep` previously bailed on the first error, aborting cleanup of all remaining resource types and discarding the run's mark state; now it collects errors, sweeps everything, persists state, and returns them as an aggregate so the job still reports the failure.

**Which issue(s) this PR fixes**:

Fixes #262

### DIFF
--- a/cmd/aws-janitor/main.go
+++ b/cmd/aws-janitor/main.go
@@ -30,6 +30,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/push"
 	"github.com/sirupsen/logrus"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
 
 	"sigs.k8s.io/prow/pkg/logrusutil"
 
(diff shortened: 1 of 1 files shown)


torredil:
cc @dims

dims:
i like this. thanks @torredil 

/approve
/lgtm

